PR #89, opened 2020-01-10
Fix README.md: Minor Corrections

danintel:
Signed-off-by: danintel <daniel.anderson@intel.com>


README.md
```diff
@@ -172,15 +172,15 @@ The two aforementioned requirements are not only satisfied by the Occlum toolcha
...
 
 ## Why the Name?
 
-The project name Occlum stems from the word *Occlumency* coined in Harry Porter series by J. K. Rowling. In *Harry Porter and the Order of Pheonix*, Occlumency is described as:
+The project name Occlum stems from the word *Occlumency* coined in Harry Potter series by J. K. Rowling. In *Harry Potter and the Order of Phoenix*, Occlumency is described as:
 
 > The magical defence of the mind against external penetration. An obscure branch of magic, but a highly useful one... Used properly, the power of Occlumency will help shield you from access or influence.
 
-The same thing can be said to Occlum, not for mind, but program:
+The same thing can be said for Occlum, not the for mind, but for the program:
```
henrysun007: I think "the" should be put after "for" right before "mind".

danintel: Yes. I added that.

tatetian:
Hi @danintel ,

Thanks for pointing out the typos. I really appreciate it.

Could you kindly rename the commit message so that it starts with "Fix"? We haven't published any contribution guideline (sorry, my fault). But the current convention is to start a commit message with a verb.

And we are about to release version 0.9.0. So if you don't like the trouble of fixing the commit message, maybe we can simply fix the typos in the new release. It is up to you.

Thanks again.

danintel:
I added "Fix" to the commit message as requested.

PR #102, opened 2020-06-04
pal: Fix crash caused by getpwuid() returning NULL

uudiin:
On lightweight Linux distribution, like alpine, getpwuid()
returns NULL, and errno is ENOENT, this patch fix crash
caused by this situation.

Signed-off-by: Tianjia Zhang <tianjia.zhang@linux.alibaba.com>


tatetian:
Hi @uudiin ,

Thanks for the bugfix. The commit looks fine to me. I will merge it shortly.

Best,
Tate

PR #103, opened 2020-06-04
pal: Fix typo error in saving token file

uudiin:
The address of the `token` should be used by fwrite().

Signed-off-by: Tianjia Zhang <tianjia.zhang@linux.alibaba.com>


tatetian:
Thanks for the bugfix! We really appreciated.

PR #115, opened 2020-07-30
Add proxy support for docker cmd

yang8621:


yang8621:
> Hi, thanks a lot for the PR! I really appreciate it.
> 
> Initially, I was 100% supportive to the approach taken in this commit. But after reading [Configure Docker to use a proxy server](https://docs.docker.com/network/proxy/), I am not so sure any more. It seems that setting environment variables within the container to enable proxy only applies to Docker 17.06 and lower. Docker 17.06, released in 2017, is a very, very old version. So the recommended method should be editing `~/.docker/config.json`, which does not involve modifying `build_image.sh`.

Thank you for the clarification. I also agree [Configure Docker to use a proxy server](https://docs.docker.com/network/proxy/) is more friendly to follow. You could balance if add any sentences to describe. Thanks.

PR #118, opened 2020-08-06
Get quote size dynamically

duanbing:


src/libos/src/util/sgx/sgx_attestation_agent.rs
```diff
@@ -133,8 +139,21 @@ impl InnerAgent {
...
+        assert!(status == sgx_status_t::SGX_SUCCESS);
+        if rt != sgx_status_t::SGX_SUCCESS {
+            match rt {
+                sgx_status_t::SGX_ERROR_BUSY => {
+                    return_errno!(EBUSY, "occlum_ocall_sgx_calc_quote_size is temporarily busy")
+                }
+                _ => return_errno!(EINVAL, "occlum_ocall_sgx_calc_quote_size failed"),
+            }
+        }
+        let mut quote_buf = vec![0_u8; quote_len as usize];
```
tatetian, line 156: An array would be  enough. Allocating a vector is more expensive. So I suggest the following line instead:
```
let mut quote_buf = [0_u8; quote_len as usize];
```

src/libos/src/util/sgx/sgx_attestation_agent.rs
```diff
@@ -133,8 +139,21 @@ impl InnerAgent {
...
-        let mut quote_buf = [0_u8; 4096];
+        let mut quote_len: u32 = 0;
+        let mut rt = Default::default();
+        let status = unsafe {
+            occlum_ocall_sgx_calc_quote_size(&mut rt as _, sigrl_ptr, sigrl_size, &mut quote_len as _)
+        };
+        assert!(status == sgx_status_t::SGX_SUCCESS);
+        if rt != sgx_status_t::SGX_SUCCESS {
+            match rt {
+                sgx_status_t::SGX_ERROR_BUSY => {
```
tatetian, line 150: According to Intel SGX SDK's manual, the only possible error is `SGX_ERROR_INVALID_PARAMETER`. So no need to check `SGX_ERROR_BUSY`.

src/libos/src/util/sgx/sgx_attestation_agent.rs
duanbing: The fixed-length array is not allowed here due to quote_len is calculated dynamically, besides, `get_quote` and size are not a frequent operation, usually just need to do for a new process, not a new task/thread.   I also have some other way to reduce the cost from heap allocation: 
1.  pre-allocate an array quote_buf[4096],  and use it when the quote_len < 4096.  Or allocate from the heap. 
2. C99 VLA-like implementation, which would be complicated and buggy.  
3. C alloca support, which allocates the buffer in the stack,  `stackvec` is an existing crate to use, but is unsafe. 

So, I think vector is a good tradeoff.

src/libos/src/util/sgx/sgx_attestation_agent.rs
duanbing: Right,  fixed.

src/libos/src/util/sgx/sgx_attestation_agent.rs
tatetian: You are right. Thanks.

tatetian:
Hi @duanbing , the code is all well.

Could you rebase the commit to the latest master (which is 0.15.0)? Then, we will be able to merge the PR. Thanks a lot!

duanbing:
@tatetian  Rebasing & Squashing done after fetching the upstream.

PR #136, opened 2020-09-17
Integrate cargo fmt with make format

jessehui:


src/exec/Makefile
```diff
@@ -17,6 +17,19 @@ else
...
+		if [[ $$retval -eq 1 ]]; then \
+			$(ECHO) "$$output"; cargo fmt; $(ECHO) "$(GREEN)\nRust code format corrected.$(NO_COLOR)"; \
+		fi
+
+format-check: $(SRC_FILES)
+	@output=$$(cargo fmt -- --check 2>&1); retval=$$?; \
+		if [[ $$retval -eq 1 ]]; then \
+			$(ECHO) "$(RED)\nRust code format error detected:$(NO_COLOR)"; $(ECHO) "\n$$output"; \
+			$(ECHO) "\nTo get rid of the format warnings above, run $(CYAN)"make format"$(NO_COLOR) to correct"; \
+		fi
```
tatetian, line 31: The format targets in `exec/Makefile` and `libos/Makefile` have some duplicated code (the format command and the warning messages). I think it is better to define functions in `sgxenv.mk` so that we do not repeat ourselves in the Makefiles in sub-projects.

Another suggestion is to change "Rust code format error detected"  to "Some format issues of Rust code are detected".

.githooks/post-commit
```diff
@@ -5,6 +5,12 @@ if ! hash astyle; then
 	exit 1
 fi
 
+cargo fmt --version >/dev/null 2>&1
+if [[ $? != 0 ]]; then
+	echo "Warning: \`cargo fmt\` is not available."
```
tatetian, line 10: I think the warning messages for missing `cargo fmt` or `astyle` should be written similarly. For the same class of warnings, we should write the warning messages in the same form.

src/libos/Makefile
```diff
@@ -136,6 +136,29 @@ $(S_OBJS):$(OBJ_DIR)/libos/$(SRC_OBJ)/%.o: src/%.S
...
+
+c_check: $(C_SRCS) $(CXX_SRCS)
+	@$(C_FORMATTER) --check $^
+
+rust_check: $(RUST_SRCS)
+	@output=$$(cargo fmt -- --check 2>&1); retval=$$?; \
+		if [[ $$retval -eq 1 ]]; then \
+			$(ECHO) "$(RED)\nRust code format error detected:$(NO_COLOR)"; $(ECHO) "\n$$output"; \
+			$(ECHO) "\nTo get rid of the format warnings above, run $(CYAN)"make format"$(NO_COLOR) to correct"; \
+		fi
```
tatetian, line 160: The naming of these targets are not consistent. Some uses `_`, while others use `-`. I suggest using the following names:
```
format
format-c
format-rust

format-check
format-check-c
format-check-rust
```

tatetian (changes requested):

tatetian:
@jessehui While it is kind of subtle, I think it is more precise to say "Integrate cargo fmt _into_ make format" (instead of "with").

src/exec/Makefile
jessehui: Done.

src/libos/Makefile
jessehui: Done.

jessehui:
> @jessehui While it is kind of subtle, I think it is more precise to say "Integrate cargo fmt _into_ make format" (instead of "with").

Done.

tatetian (approved):
The commit looks perfect. Thanks a lot.

PR #155, opened 2020-09-22
Add Golang installer

jessehui:


tools/installer/README.md
```diff
@@ -99,7 +107,7 @@ If a user wants to build his application on a platform installed with Occlum ins
 cd tools/installer/deb
 make <language option>
 ```
-Now, only `c/c++` option is supported. And the installer can be found under `build/debs`.
+Now, only `c/c++` and `golang` option is supported. And the installer can be found under `build/debs`.
```
tatetian, line 110: One piece of general writing tips:
> A proper noun is the special noun or name used for a specific person, place, company, or other thing. Proper nouns should always be capitalized.

Both C and C++ are proper nouns; so they should be capitalized. So is Go. When writing docs (rather than code), the capitalization rule should be followed. I hope this tips can make you more correct, consistent, and confident when dealing with capitalization in the future :)

Another issue is that almost all occurrences of `` `C/C++` `` and `` `golang` `` in the modifications of this commit should not be surrounded by backticks ``` `` ```: they are NOT code, but proper nouns.

jessehui: Got it. Thank you for your review. I use \` \` because it is the option provided with `make`.  For user options, I think using lower-case is more appropriate. Thus, I will keep \`\` and lower-case for user option and change others to upper-case.

guzongmin (approved):
LGTM

PR #164, opened 2020-09-24
Fix copy softlink for make install and deb installer

jessehui:
man cp：
> -a, --archive
same as -dR --preserve=all
-d
same as --no-dereference --preserve=links
-P, --no-dereference
never follow symbolic links in SOURCE
--preserve[=ATTR_LIST]
preserve the specified attributes (default: mode,ownership,timestamps), if possible additional attributes: context, links, xattr, all


guzongmin (approved):
LGTM

PR #165, opened 2020-09-24
Add detection for FSGSBASE enablement in PAL

jessehui:
Now, we are in a complex situation that FSGSBASE instructions might or might not be supported in the OS. Add this new tool for users to do a quick check on whether it is supported.

In the future, we can add more checks and diagnose in this subcommand to include more info that we are interested in. Thus, we can let users run this command if they can't run Occlum successfully. And If they can't fix for themselves, they can paste the log to GitHub issues and it would be easier for us to diagnose for them.

### DIFF
--- a/src/pal/src/pal_check_fsgsbase.c
+++ b/src/pal/src/pal_check_fsgsbase.c
@@ -1,0 +1,63 @@
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <signal.h>
+#include <assert.h>
+#include <string.h>
+#include <ucontext.h>
+#include <setjmp.h>
+#include <stdlib.h>
+#include <errno.h>
+
+#include "pal_log.h"
+#include "pal_check_fsgsbase.h"
+
+#define RC 0xffff
+static jmp_buf env_buf;
+
+static void handle_sigill(int num) {
+    assert(num == SIGILL);
+
+    longjmp(env_buf, RC);
+}
+
+int check_fsgsbase_enablement(void) {
+    int gs_read_data = 0;
+    int gs_write_data = 0x0f;
+    int __seg_gs *offset_ptr = 0;   // offset relative to GS. support since gcc-6
+
+    sighandler_t handler_orig = signal(SIGILL, handle_sigill);
+    if (handler_orig == SIG_ERR) {
+        PAL_ERROR("registering signal handler failed, errno = %d", errno);
+        return errno;
+    }
+
+    int ret = setjmp(env_buf);
+    if (ret == RC) {
+        // return from SIGILL handler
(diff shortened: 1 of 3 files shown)


jessehui:
The result will look like this:
截屏2020-09-24下午8 50 54

截屏2020-09-24下午8 50 37

tools/occlum
```diff
@@ -443,6 +476,9 @@ case "$cmd" in
     status)
         cmd_status
         ;;
+    check)
+        cmd_check
```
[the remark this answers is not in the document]
tatetian: Agree with @guzongmin . Useful feature, but not important enough to qualify as a command.

Instead of asking users to learn a new command, I think a more desired user experience is to do this check behind the scene automatically: before every `occlum run` or `occlum exec` command, run the `detect_fsgsbase` tool. If found any problem, output error messages to the user. If it is ok, save the success result to some location so that the tool does not need to be run again for the same Occlum instance.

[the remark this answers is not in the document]
jessehui: Thank you for your review. Move this check to `occlum run/start`

tools/detect_fsgsbase/detect_fsgsbase.c
```diff
@@ -0,0 +1,48 @@
...
+static void handle_sigill(int num) {
+    printf("\tSIGILL Caught !\n");
+    assert(num == SIGILL);
+    printf("\tFSGSBASE instructions are not supported.\n");
+
+    // since we don't know how far we should jump, just use longjmp to exit
+    longjmp(env_buf, RC);
+}
+
+int main (void) {
```
guzongmin, line 23: Since the code is very simple, should we add it in the occlum_pal.so just before launch the occlum instance?

tatetian: Agree with @guzongmin . Let's add it to PAL.

src/pal/src/pal_check_fsgsbase.c
```diff
@@ -0,0 +1,58 @@
...
+    asm("wrgsbase %0" :: "r" (&gs_write_data));
+
+    if (*offset_ptr != 0x0f) {
+        PAL_ERROR("GS register data not match\n");
+        return -1;
+    };
+
+    // restore the GS register
+    asm("wrgsbase %0" :: "r" (&gs_read_data));
+
```
guzongmin, line 56: signal(SIGILL, SIG_DFL);

Please restore the default handler.

jessehui: Done. Thank you for your review.

guzongmin (approved):
LGTM

PR #168, opened 2020-09-27
Enhance Socket

henrysun007:
1. Add Rust memory-safe types;
2. Add Berkeley Sockets API and implement it for Socket.

Cross-world UDS needs this commit.


src/libos/src/net/socket/socket_file.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+        Ok(total_len)
+    }
+
+    fn writev(&self, bufs: &[&[u8]]) -> Result<usize> {
+        let mut total_len = 0;
+        for buf in bufs {
+            match self.write(buf) {
+                Ok(len) => {
+                    total_len += len;
```
tatetian, line 53: What if `len < buf.len()`? The code should be
```
Ok(len) => {
    total_len += len;
    if len < buf.len() {
        break;
    }
}
```

src/libos/src/net/socket/socket_file.rs
```diff
@@ -0,0 +1,97 @@
...
+        self.write(buf)
+    }
+
+    // TODO: use sendmsg to impl readv
+    fn readv(&self, bufs: &mut [&mut [u8]]) -> Result<usize> {
+        let mut total_len = 0;
+        for buf in bufs {
+            match self.read(buf) {
+                Ok(len) => {
+                    total_len += len;
```
tatetian, line 39: What if `len < buf.len()`? The code should be
```
Ok(len) => {
    total_len += len;
    if len < buf.len() {
        break;
    }
}
```

src/libos/src/net/socket/socket_file.rs
```diff
@@ -0,0 +1,97 @@
...
+                }
+                Err(_) if total_len != 0 => break,
+                Err(e) => return Err(e.into()),
```
tatetian, line 42: I think `Err(e.into())` can be simplified to `Err(e)`.

src/libos/src/net/socket/address_family.rs
```diff
@@ -0,0 +1,71 @@
...
+    IEEE802154 = 36,
+    CAIF = 37,
+    ALG = 38,
+    NFC = 39,
+    VSOCK = 40,
+    KCM = 41,
+    QIPCRTR = 42,
+    SMC = 43,
+    XDP = 44,
+    MAX = 45,
```
tatetian, line 60: Is `AddressFamily::MAX` a valid value for address family? If it is not, it should not be an enum value. Unlike C, Rust has a very strong and expressive type system. An object of a type must have a proper value of the type. A const value can be defined for your purpose here.

tatetian:
@henrysun007 By the way, I think the commit message is a bit of short and vague. Could you improve it?

src/libos/src/net/socket/berkeley_sockets.rs
```diff
@@ -0,0 +1,14 @@
+use super::*;
+
+// Trait for Berkeley socket API
```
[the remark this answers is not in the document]
henrysun007: I am also confused about that `s`, but it is used everywhere. Please google `bsd sockets api` and you can see yourself.

src/libos/src/net/socket/address_family.rs
henrysun007: Yes. There is a kind of family called `AF_MAX`. Refer to https://github.com/ifduyue/musl/blob/33338ebc853d37c80f0f236cc7a92cb0acc6aace/include/sys/socket.h#L197.

src/libos/src/net/socket/socket_file.rs
henrysun007: I used sendmsg/recvmsg to implement readv/writev.

src/libos/src/net/socket/berkeley_sockets.rs
```diff
@@ -0,0 +1,14 @@
...
+// TODO: add more functions
+pub trait BSDSockets {
+    fn bind(&self, addr: SockAddr) -> Result<()>;
```
tatetian, line 6: The argument `addr` should be a reference to prevent the caller of this method from losing the ownership of `addr` object. The reasoning applies to the `connect` and `sendto` method.

src/libos/src/net/socket/socket_address.rs
```diff
@@ -0,0 +1,121 @@
...
+            .finish()
+    }
+}
+
+impl SockAddr {
+    // Caller should guarentee the sockaddr and addr_len are valid
+    pub unsafe fn try_from_raw(
+        sockaddr: *const libc::sockaddr,
+        addr_len: libc::socklen_t,
+    ) -> Result<Self> {
```
tatetian, line 28: I cannot find any code that uses this method. Could you explain?

henrysun007: It is used to transform `libc::sockaddr` to Rust `SockAddr`. You can find the usage in `src/libos/src/net/syscalls.rs`.

tatetian: Ok. I see. Thanks.

src/libos/src/net/socket/socket.rs
```diff
@@ -0,0 +1,14 @@
+use super::*;
+
+// Trait for Berkeley sockets API
+// TODO: add more functions
+pub trait Socket {
+    fn bind(&self, addr: &SockAddr) -> Result<()>;
+    fn listen(&self, backlog: i32) -> Result<()>;
+    fn accept(&self, flags: FileFlags) -> Result<(Self, Option<SockAddr>)>
```
tatetian, line 8: Why returning `Option<SockAddr>`? I think an address can always be returned from `accept`.

src/libos/src/net/syscalls.rs
```diff
@@ -80,93 +118,55 @@ pub fn do_accept4(
...
-    let need_check: bool = !addr.is_null();
-
-    if addr.is_null() ^ addr_len.is_null() {
-        return_errno!(EINVAL, "addr and ddr_len should be both null");
-    }
-    if need_check {
+    let addr_set: bool = !addr.is_null();
+    if addr_set {
+        from_user::check_ptr(addr_len)?;
         from_user::check_mut_array(addr as *mut u8, unsafe { *addr_len } as usize)?;
```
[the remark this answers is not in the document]
henrysun007: `addr_len` is checked just in the last line.

src/libos/src/net/syscalls.rs
```diff
@@ -80,93 +118,55 @@ pub fn do_accept4(
     addr_len: *mut libc::socklen_t,
     flags: c_int,
 ) -> Result<isize> {
-    debug!(
-        "accept4: fd: {}, addr: {:?}, addr_len: {:?}, flags: {:#x}",
-        fd, addr, addr_len, flags
-    );
```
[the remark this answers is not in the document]
henrysun007: The raw inputs have already been printed in the entry of syscall.

src/libos/src/net/socket/socket.rs
henrysun007: It happens when no address is returned. The POSIX API returns zero-length.

src/libos/src/net/socket/socket.rs
```diff
@@ -0,0 +1,14 @@
+use super::*;
+
+// Trait for Berkeley sockets API
+// TODO: add more functions
+pub trait Socket {
```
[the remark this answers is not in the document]
henrysun007: Agree. We will see the benefit of this unified trait with `cross-world unix socket`.

src/libos/src/net/socket/socket_address.rs
```diff
@@ -0,0 +1,121 @@
+use super::*;
+use std::*;
+
+#[derive(Copy, Clone)]
+pub struct SockAddr {
+    storage: libc::sockaddr_storage,
+    len: usize,
+}
```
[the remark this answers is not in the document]
henrysun007: Agree. This is a little trick to not to maintain different address types. There are too many address types. For IP socket, this structure is enough. I add checks before converting C types to Rust types. For UDS, there will be a specific address type.

tatetian (approved):
LGTM

PR #169, opened 2020-09-28
Get rid of occlum-gen-default-occlum-json script and python dependency

jessehui:
Integrate `occlum-gen-default-occlum-json` script with the previous `gen_enclave_conf` and rename to `gen_internal_conf` which are now used generate both internal Occlum.json and Enclave.xml. 
Also, remove python function from `occlum` script.


tools/occlum_build.mk
```diff
@@ -28,36 +28,11 @@ define get_conf_root_fs_mac
...
-	cat "$(JSON_CONF)" | \
-		python -c "import sys, json; print json.load(sys.stdin)['process']['default_heap_size']"
-endef
-
-define get_conf_default_mmap_size
-	cat "$(JSON_CONF)" | \
-		python -c "import sys, json; print json.load(sys.stdin)['process']['default_mmap_size']" ['resource_limits']['user_space_size']
-endef
-
 define get_conf_user_space_size
```
guzongmin, line 31: Do we need it?

tools/gen_internal_conf/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+
+fn main() {
+    env_logger::init();
+
+    let matches = App::new("gen_internal_conf")
+        .version("0.2.0")
+        // Input: JSON file which users may change
+        .arg(
+            Arg::with_name("user_json")
+                .short("i")
```
guzongmin, line 29: The short name and the long name does not match.

tools/occlum_build.mk
jessehui: Removed

tools/gen_internal_conf/src/main.rs
jessehui: short name removed.

tools/gen_internal_conf/src/main.rs
```diff
@@ -0,0 +1,355 @@
+extern crate clap;
```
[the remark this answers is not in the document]
jessehui: Thank you for your review. Parsing can't be done if you don't tell the parser what new content to parse. We don't know what the JSON would be like in the future. Maybe a whole section will be added or removed. To handle all situations, a lot of effort is needed. And still, it may not work when we really change the JSON. 
This commit already eliminated the effort to modify the options in the JSON file. We may be able to find a way to do it really automatically, but currently, the most important thing is to find a balance between the effort and the effect, right?

guzongmin (approved):
LGTM

PR #171, opened 2020-10-06
Add Unit Testing framework

henrysun007:


guzongmin (commented):
I have concern about how it works to handle some error cases which may make the LibOS crash. UT needs to test those corner cases which may never triggered by normal test cases. Most of those test case may crash the LibOS (LibOS must crash to protect the security of user's data).

henrysun007:
@guzongmin I added support for `#[should_panic]`. Is it what you want?

src/libos/src/unit_test/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+# Run unit_test along with other tests:
+$ make test OCCLUM_UT=1
+
+# Only run unit_test
+$ make test TESTS=unit_test
+
+# Assign the test_name, like net::test::test_iov, to BIN_ARGS to run specific test
+$ make test BIN_ARGS=<test_name> TESTS=unit_test
+```
```
tatetian, line 27: I think the doc should be put to where it would be most likely to be used.

For example, when running unit tests, a developer would most likely to try to find relevant info in the `test` directory, rather than the `src` directory. Actually, mentioning `make test` and its usage in source code itself is not a good idea, which entangles the source code and the test infrastructure.

src/libos/src/unit_test/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Adding a new test
+Define a new function with the type `fn() -> ()` and add `#[occlum_test]` attribute for it.
+Rust built-in attribute `#[should_panic]` without optional parameter is also supported.
```
tatetian, line 31: This should be written as Rust comments, instead of as part of README.

src/libos/src/unit_test/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+
+use super::*;
+use std::panic;
+use std::sync::atomic::{AtomicIsize, Ordering};
+use util::mem_util::*;
+
+static NUM_PASS: AtomicIsize = AtomicIsize::new(0);
+static NUM_FAIL: AtomicIsize = AtomicIsize::new(0);
+
+const GREEN_OK: &str = "\x1b[0;32mok\x1b[0m";
```
tatetian, line 11: Use lowercase "ok" but uppercase "FAILED"?

src/libos/src/unit_test/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+const RED_FAILED: &str = "\x1b[0;31mFAILED\x1b[0m";
+
+pub fn unit_test(cmd: usize, func_name_ptr: *const c_char, index: i32) -> Result<isize> {
```
tatetian, line 14: I think the interface of this `unit_test` syscall can be improved to the following form:
```
pub fn unit_test(test_prefix: *const c_char, int *total_count, int *pass_count) -> Result<isize>
```
with the arguments
* `test_prefix` is an optional input, specifying the prefix of the unit tests should be run. If not given, all test cases should be run.
* `total_count` is an output, giving the total number of test cases that run.
* `pass_count` is an output, giving the number of test cases that pass.

The new interface is not only more powerful (allowing prefix/exact matching), but also simplifies the user and kernel code 
* The user program that drives the unit tests (`test/unit_test/main.c`) can be greatly simplified;
* The user program that runs individual test case (`test/syscall/main.c`) is no longer needed;
* The kernel code does not need to maintain state (e.g., `NUM_PASS` or `NUM_FAIL`) and does not need to have separate commands (`GetCount`, `GetPass`, `RunByName`, and `RunByIndex`);
* Each unit test does need to be run in separate syscall (I cannot see why this is more desirable).

src/libos/src/untrusted/test.rs
```diff
@@ -0,0 +1,18 @@
...
+    assert_eq!(Some(empty_slice).as_ptr_and_len(), (std::ptr::null(), 0));
+}
+
+#[occlum_test]
+pub fn test_none_slice_ext() {
+    assert_eq!(
+        Option::<&[u8]>::None.as_ptr_and_len(),
+        (std::ptr::null(), 0)
+    );
+}
```
tatetian, line 18: Unit tests need to be put close to their target code.

src/libos/src/unit_test/test.rs
```diff
@@ -0,0 +1,7 @@
+use super::*;
+
+#[occlum_test]
+#[should_panic]
+pub fn test_should_panic() {
+    panic!();
+}
```
tatetian, line 7: This is too trivial to be a separate file.

Can we have a meaningful `should_panic` unit test somewhere? Or this example can be added as a code snippet to the module-level Rust doc of `unit_test` module.

src/libos/src/unit_test/proc_macro/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0.24"
+quote = "1.0.7"
+syn = { version = "1.0.42", features = ["parsing"] }
+inventory = "0.1.9"
```
tatetian, line 13: Having a separate crate under `libos/src` breaks the current unwritten convention of code organization. Having a crate embedded in the directory of another crate does not seem to be a good practice for Rust projects. The best solution that comes to my mind right now is to add a `crates` directory under `libos` directory. This approach is clear and extensible: more crates can be added in the future.

Could you try rewriting the `unit_test` module and the `proc_macro` crate as separate crates under the `crates` directory? Also, can we combine them into one crate?

src/libos/Makefile
```diff
@@ -106,10 +106,18 @@ $(LIBOS_CORE_A): $(LIBOS_CORE_RS_A) $(C_OBJS) $(CXX_OBJS) $(S_OBJS) $(EDL_C_OBJS
 	@ar r $@ $(C_OBJS) $(CXX_OBJS) $(S_OBJS) $(EDL_C_OBJS)
 	@echo "GEN => $@"
 
+ifeq ($(OCCLUM_UT), 1)
+    UT_FLAGS := --features "unit_test"
+    # inventory does not work in debug mode
```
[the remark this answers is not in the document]
henrysun007: The  issue is track in https://github.com/dtolnay/inventory/issues/24.

src/libos/src/unit_test/mod.rs
henrysun007: Yes. I just keep the output consistent with `cargo test`. Below is the output of a failed `cargo test`:
截屏2020-10-18 下午8 09 41

src/libos/src/unit_test/test.rs
henrysun007: Basically, this is a test for `unit_test`. I put it close to `unit_test`.

src/libos/src/unit_test/proc_macro/Cargo.toml
henrysun007: I added `proc_macro` in the crates directory. It has no dependencies on occlum. Instead, `unit_test` module relies on occlum heavily. I leave the modularization for future work.

src/libos/src/unit_test/mod.rs
henrysun007: I removed `test/syscall`.

src/libos/src/unit_test/README.md
henrysun007: Developers can find all they want to know about occlum unit test in one file. I can put the contents to different places if you have any concrete idea.

src/libos/src/unit_test/mod.rs
tatetian: Got it. Thanks.

src/libos/Makefile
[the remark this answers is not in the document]
tatetian: Thanks for pointing out that the `inventory` crate is not working for incremental builds. But I am afraid that this problem needs to be fixed before this PR can be accepted. The reason is as follows.

As the release build of a Rust project is painfully slow, almost all development (unless it is performance related) would be carried out under debug builds. And it is the development phase where unit tests are most useful. So, if unit tests cannot work for debug builds, then their value would be greatly diminished.

src/libos/src/unit_test/test.rs
tatetian: Ok

src/libos/src/unit_testing/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::panic;
+use std::sync::atomic::{AtomicIsize, Ordering};
+use util::mem_util::*;
+
+const GREEN_OK: &str = "\x1b[0;32mok\x1b[0m";
+const RED_FAILED: &str = "\x1b[0;31mFAILED\x1b[0m";
+
+pub fn unit_testing(name_prefix: *const c_char) -> Result<isize> {
+    if name_prefix.is_null() {
+        let tests = inventory::iter::<TestCase>.into_iter().collect::<Vec<_>>();
```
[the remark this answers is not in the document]
henrysun007: It is used to fit for `run_tests`.

src/libos/src/crates/test_proc_macro/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+        #[cfg(feature = "unit_testing")]
+        #func
+
+        #[cfg(feature = "unit_testing")]
+        inventory::submit!(
+            unit_testing::TestCase::new(
+                module_path!().replace("occlum_libos_core_rs::", "").to_string() +
+                     "::" + stringify!(#func_ident),
+                #func_ident, #should_panic,
+            )
```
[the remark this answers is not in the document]
henrysun007: Procedural macros must be defined in a crate with the crate type of proc-macro. Refer to https://doc.rust-lang.org/reference/procedural-macros.html for more information.

tatetian:
@henrysun007 Before checking your updates to this PR , I hope to get some comments from you about the approach to unit testing for no_std environments described by [this blog post](https://os.phil-opp.com/testing/). It is based on Rust's [custom test frameworks feature](https://github.com/rust-lang/rust/issues/50297). From the two sources of information, I got the impression that the approach has the benefits: 1) it does not require writing a crate, and 2) it works for incremental builds.  From this perspective, the approach described in the blog post seems to be superior to the one proposed by this PR.

henrysun007:
We talked about this framework the first time I intended to add UT. That framework has the support of cargo, which means it also is restricted by cargo. Cargo will output an executable with the tests. As far as I know, Occlum cannot run as an executable with minor modification.

tatetian:
> We talked about this framework the first time I intended to add UT. That framework has the support of cargo, which means it also is restricted by cargo. Cargo will output an executable with the tests. As far as I know, Occlum cannot run as an executable with minor modification.

While the blog test does use `cargo test`, it does not mean it is impossible to use the test framework without `cargo test`. I think what is important is that the framework helps collecting all test cases at compile time. With the test cases collected, we can take care of running them by ourselves.

PR #172, opened 2020-10-08
Refactor pipe with the new event subsystem

tatetian:
1. Introduce channels, which provide an efficient means for IPC;
2. Leverage channels to rewrite pipe, improving the performance (3X), robustness, and readability.

This pipe rewrite is not done: some more commits will be added to implement poll and epoll for pipe.

### DIFF
--- a/src/libos/Cargo.lock
+++ b/src/libos/Cargo.lock
@@ -5,6 +5,7 @@ name = "Occlum"
 version = "0.16.0"
 dependencies = [
  "aligned",
+ "atomic",
  "bitflags",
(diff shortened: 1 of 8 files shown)


tatetian:
The failed demo CI "C_cpp_rust_golang_embedded_mode_support_test" is due to the known limitation that polling on the new implementation of pipe is not implemented, yet. I think it is ok to accept this commit even though it makes some tests fail. I am confident that the refactoring of pipe will be complete before the next release.

guzongmin (approved):
LGTM

PR #177, opened 2020-10-10
Update issue templates

guzongmin:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
+---
```
henrysun007, line 1: Below is the bug report template from aone:
[缺陷描述]：
[重现步骤]：
[期望结果]：
[实际结果]：
[原因定位]：
[修复建议]： 
It has the `root cause` that is missing in this commit. Besides, should `Environment` and `Additional context` be in `To Reproduce`?

guzongmin: It is bug report template. Most of the user do not know what is the root cause. So I do not add it in the template.
"To Reproduce" highlight the actions of user. It should be short and clear.
"Enviroment" is a hint for user to check what is the execution enviroment. Most of the user would forget tell us those infomation, so I add it to highlight it.
"Additional context" would be anything useful. For example some other softwares are installed on the machine has similar issue too. But user does not know the relationship between them. Just tell us for reference.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
...
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: ''
+assignees: guzongmin
+
+---
+
+<!--- Provide a general summary of the issue in the Title above -->
+**Describe the bug**
```
tatetian, line 11: Instead of formatting the title as below,
````
**Describe the bug**
````
I think it is more appropriate to format it as the title of a section
````
# Describe the bug
````
This way, the submitter of the issue can have subsections or subsubsections to make the text look more structured.

File: .github/ISSUE_TEMPLATE/bug_report.md [hunk 3]
```diff
@@ -0,0 +1,36 @@
...
+
+**Environment (please complete the following information):**
+- HW: [e.g. SGX1, SGX2] 
+- OS: [e.g. Ubuntu18.04, CentOS8.1]
+ - Occlum version [e.g. 0.16.0]
+
+**Additional context**
+Add any other context about the problem here.
+
+**Possible Solution/Implementation**
```
tatetian, line 35: The titles are not capitalized consistently. Some are capitalized (e.g., `To Reproduce`), while some do not seem to be capitalized (such as `Additional context`). I think it is ok to choose not capitalizing titles in informal writing. But one should not capitalize some tiles, while ignoring the others. Make a choice about the style and stick to it.

If you decide to capitalize the titles, I recommend using [this online tool](https://capitalizemytitle.com/) to automatize the task.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
...
+3. 
+4. 
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Logs**
+If applicable, add logs to help explain your problem.
+
+**Environment (please complete the following information):**
```
tatetian, line 27: I think `(please complete the following information)` should be removed.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
...
+- HW: [e.g. SGX1, SGX2] 
+- OS: [e.g. Ubuntu18.04, CentOS8.1]
+ - Occlum version [e.g. 0.16.0]
```
tatetian, line 30: Remove the whitespace at the beginning of this line.

.github/ISSUE_TEMPLATE/bug_report.md
guzongmin: It is used to highlight that the below information is important.

tatetian: Maybe it needs some rewording. The sentence looks redundant to me. I think every section requires the issue submitter to complete some information, doesn't it? Why need to point that out for the environment section in particular?

tatetian (approved):
LGTM

PR #178, opened 2020-10-10
Update the xgboost to latest version

yuanwu2017:
Intel has optimized the xgboost in latest version. Performance has
obvious acceleration in benchmark. In svm training with Yahoo LTR data,
latest version gets 7.8X acceleration in host, and 2X acceleration in occlum.

Signed-off-by: yuanwu <yuan.wu@intel.com>


guzongmin (approved):
LGTM

Issue #173, opened 2020-10-09
Golang demo in CI failed


henrysun007:
Error message:
```
Run docker exec language_support_test bash -c "cd /root/occlum/demos/golang/grpc_pingpong && ./prepare_ping_pong.sh"
go: finding google.golang.org/protobuf v1.25.0
go: downloading google.golang.org/protobuf v1.25.0
go: extracting google.golang.org/protobuf v1.25.0
go: finding google.golang.org/grpc v1.32.0
go: downloading google.golang.org/grpc v1.32.0
go: finding google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.0.0
go: downloading google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.0.0
go: extracting google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.0.0
go: extracting google.golang.org/grpc v1.32.0
go: downloading google.golang.org/grpc v1.33.0-dev
go: extracting google.golang.org/grpc v1.33.0-dev
go: downloading golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a
go: downloading github.com/golang/protobuf v1.4.1
go: downloading golang.org/x/net v0.0.0-20190311183353-d8887717615a
go: downloading google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
go: extracting github.com/golang/protobuf v1.4.1
go: extracting golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a
go: extracting golang.org/x/net v0.0.0-20190311183353-d8887717615a
go: downloading golang.org/x/text v0.3.0
go: extracting golang.org/x/text v0.3.0
go: extracting google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
go: finding google.golang.org/grpc v1.33.0-dev
go: finding golang.org/x/net v0.0.0-20190311183353-d8887717615a
go: finding github.com/golang/protobuf v1.4.1
go: finding google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
go: finding golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a
go: finding golang.org/x/text v0.3.0
# command-line-arguments
./pong.go:32:2: undefined: pingpong.RegisterPingPongServiceService
./pong.go:32:55: undefined: pingpong.PingPongServiceService
Error: Process completed with exit code 2.
```

jessehui:
[...]

PR #180, opened 2020-10-10
Fix an issue caused by recent reverted gRPC service registration interfaces

james-dong-security:
in cmd/protoc-gen-go-grpc.

This patch will resolve issue https://github.com/occlum/occlum/issues/173.


guzongmin (approved):
LGTM

PR #182, opened 2020-10-10
add font support for occlum libos

superajun-wsj:


guzongmin (commented):
Two general comments:
1. There are too many binaries in this patch. Can you please provide a script to download them instead of commit to git?
2. The folder deps is used for dependencies used to build Occlum. The font supports should be in tools folder.

superajun-wsj:
font support for occlum will be merge in other pull request: https://github.com/occlum/occlum/pull/186, so close this one.

PR #190, opened 2020-10-19
Put the shebang in the front of run_occlum_bench.sh

henrysun007:
Exec parses the script according to the shebang.


guzongmin (approved):
LGTM

PR #192, opened 2020-10-21
Fix package build error

jessehui:
1. Fix repeatedly linking dynamic libraries for occlum-run
2. Fix gcc toolchain deb package build error
3. Remove redundant file from installer file list
4. Change symlink target path to relative path

### DIFF
--- a/tools/installer/deb/toolchains/gcc/rules
+++ b/tools/installer/deb/toolchains/gcc/rules
@@ -15,6 +15,12 @@ MUSL_REPO = https://github.com/occlum/musl
 MUSL_VER = 1.1.24
 GCC_VER = 8.3.0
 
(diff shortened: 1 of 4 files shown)


tools/installer/deb/toolchains/gcc/rules
```diff
@@ -15,6 +15,8 @@ MUSL_REPO = https://github.com/occlum/musl
 MUSL_VER = 1.1.24
 GCC_VER = 8.3.0
 
+BUILD_OPTIONS +=
```
guzongmin, line 18: Please add a hint for user about how to use the BUILD_OPTIONS.

guzongmin (approved):
LGTM

PR #194, opened 2020-10-22
Fix the CI test failure

henrysun007:
There is a CI test failure related to `server_epoll` test. These two commits help to mitigate and diagnose the failure.


test/server_epoll/main.c
```diff
@@ -123,9 +123,13 @@ int test_ip_socket() {
                 }
             } else if (events[i].events & EPOLLIN) {
                 // Channel is ready to read.
-                char buf[36];
+                int buf_len = 36;
```
guzongmin, line 126: Should we define buf_len as a MACRO?

guzongmin (approved):
LGTM

PR #200, opened 2020-10-27
Add doc about using occlum package repo

jessehui:


docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
...
+Please follow [Intel SGX Installation Guide](https://download.01.org/intel-sgx/sgx-linux/2.11/docs/Intel_SGX_Installation_Guide_Linux_2.11_Open_Source.pdf) to install SGX driver and SGX PSW. SGX SDK is not required. Using RPM installer is recommanded. 
+
+Also, UAE service libraries are needed but may not installed together with SGX PSW if SGX PSW installer is used. Go to SGX RPM local repo and run:
+```
+rpm -i libsgx-uae-service*.rpm
+```
+
+After installing SGX driver and PSW, please make sure `aesm` service is running by checking:
+```
+service aesmd status
```
guzongmin, line 31: Should we add the expected result of command "service aesmd status" here?

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
...
+echo "source /etc/profile" >> $HOME/.bashrc
+```
+
+
+### Install Occlum with APT
+
+1. Install Prerequisite
+```
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends ca-certificates gnupg2 jq make gdb wget libfuse-dev libtool
```
guzongmin, line 57: If user only want to run the occlum instance, user does not need all those dependents. Should we separate this list to two?

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+3. Install Occlum
+```
+echo 'deb [arch=amd64] https://occlum.io/occlum-package-repos/debian bionic main' | tee /etc/apt/sources.list.d/occlum.list
+wget -qO - https://occlum.io/occlum-package-repos/debian/public.key | apt-key add -
+apt-get update
+apt-cache policy occlum
+apt-get install -y occlum
+echo "source /etc/profile" >> $HOME/.bashrc
```
tatetian, line 82: I don't quite understand this line.

First, `/etc/profile` is the system-wide environment file that is used by all users and all shells (e.g., bash) on a system. On the other hand, `~/.bashrc` is the per-user config for bash. This means whatever inside the `/etc/profile` should have been applied to the current user's bash shell; no need to add it to `~/.bashrc` by hand.

Second, I am not aware of any Occlum's modifications to `/etc/profile`. So I don't know why we need to `source /etc/profile`.

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
+# Install Occlum Packages with Package Mangement Tool
+
+Occlum packages can be easily installed and deployed using mainstream package management tool including RPM and APT.
+
+
+### Prerequisite
```
tatetian, line 6: Why go straight from `# Heading level 1` to `### Heading level 3`, skipping `# Heading level 2`?

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
+# Install Occlum Packages with Package Mangement Tool
+
+Occlum packages can be easily installed and deployed using mainstream package management tool including RPM and APT.
```
tatetian, line 3: Writing suggestions:
```
# Install Occlum with Popular Package Managers

Occlum can be easily installed with popular package managers like [APT](https://en.wikipedia.org/wiki/APT_(software)) and [RPM](https://en.wikipedia.org/wiki/RPM_Package_Manager). This document walks you through the steps to install Occlum on popular Linux distributions like Ubuntu and CentOS using package managers.
```

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
...
+```
+service aesmd status
+```
+
+3. Install Occlum
+```
+echo 'deb [arch=amd64] https://occlum.io/occlum-package-repos/debian bionic main' | tee /etc/apt/sources.list.d/occlum.list
+wget -qO - https://occlum.io/occlum-package-repos/debian/public.key | apt-key add -
+apt-get update
+apt-cache policy occlum
```
tatetian, line 80: The purpose of this command seems to be a check to see whether the Occlum APT repo is installed on the machine. If so, I think this is not necessary here.

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+After installing PSW, please make sure that the aesm service is running:
+```
+service aesmd status
+```
+
+3. Install Occlum
+```
+echo 'deb [arch=amd64] https://occlum.io/occlum-package-repos/debian bionic main' | tee /etc/apt/sources.list.d/occlum.list
```
tatetian, line 77: The repo is only applicable to Ubuntu 18.04. This restriction needs to be mentioned in advance. The same argument applies to RPM and CentOS.

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
...
+apt-get install -y occlum
+echo "source /etc/profile" >> $HOME/.bashrc
+```
+
+
+### Hello World Test
+
+You are all set. Now, let's run a hello world test:
+```
+cd $HOME && wget https://raw.githubusercontent.com/occlum/occlum/master/demos/hello_c/hello_world.c
```
tatetian, line 90: `$HOME` is not a good choice. First, people will wonder if there are some implicit requirements (which is not the case) that the commands must be executed under `$HOME`. Second, choosing `$HOME` may overwrite existing files on the user's machine accidentally or fail due to there is already a directory named `occlum-instance`.

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
...
+
+Occlum packages can be easily installed and deployed using mainstream package management tool including RPM and APT.
+
+
+### Prerequisite
+
+**Install enable_RDFSBASE Kernel Module**
+
+Please follow [this README](https://github.com/occlum/enable_rdfsbase/blob/master/README.md) to install `enable_rdfsbase` kernel module.
+
```
tatetian, line 11: Need to mention the kernel module is not needed for the latest version of Linux kernel (e.g., 5.9)

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,99 @@
...
+### Install Occlum with RPM
+
+1. Install Prerequisite
+```
+yum install -y wget yum-utils make jq gdb
+```
+
+2. Install Intel® SGX Driver and Intel® SGX PSW
+
+Please follow [Intel SGX Installation Guide](https://download.01.org/intel-sgx/sgx-linux/2.11/docs/Intel_SGX_Installation_Guide_Linux_2.11_Open_Source.pdf) to install SGX driver and SGX PSW. SGX SDK is not required. Using RPM installer is recommanded.
```
tatetian, line 22: Should we refer the doc or we can just extract the commands from the doc to save our readers from looking into the doc themselves? The latter one seems more user friendly to me. I think it is acceptable since Occlum has some requirements on the version of Intel SGX PSW. So the installation steps of that version of SGX PSW won't be updated by Intel in the future and we can write them down here.

docs/install_occlum_packages.md
jessehui: Done.

docs/install_occlum_packages.md
jessehui: Done.

docs/install_occlum_packages.md
jessehui: Done.

docs/install_occlum_packages.md
jessehui: True. Removed.

docs/install_occlum_packages.md
jessehui: Occlum installers need to export some environment variables and those are done by a script in /etc/profile.d.
It's true that when the system start, the script in /etc/profile.d will all be executed. But some servers may never be shut down and the user newly login can get occlum command without reboot the system. Thus, put that to .bashrc seems a safer way.

docs/install_occlum_packages.md
jessehui: Got it. Changed to /tmp.

docs/install_occlum_packages.md
jessehui: Besides PSW, there is also SGX driver. I agree that put installation guide here can be convenient for users. However, installing driver and PSW is not that easy. And users can meet all kinds of problems. I would hand in the professional stuff to professional people. Thus, I choose to refer to the original documents. I think PSW version should be forward compatible and thus should not cause problems.

jessehui (commented):
Thanks for your review.

docs/install_occlum_packages.md
jessehui: I think these are the minimal dependencies. wget certificate and gnupg are used for package installation. Others are needed by occlum commands.

jessehui:
Add DCAP packages.

jessehui:
Can we merge this in the next release?

docs/install_occlum_packages.md
```diff
@@ -0,0 +1,102 @@
...
+Occlum can be easily installed with popular package managers like [APT](https://en.wikipedia.org/wiki/APT_(software)) and [RPM](https://en.wikipedia.org/wiki/RPM_Package_Manager). This document walks you through the steps to install Occlum on popular Linux distributions like Ubuntu and CentOS using package managers.
+
+## Prerequisite
+
+**Install enable_RDFSBASE Kernel Module**
+
+If the Kernel version is before `v5.9`, please follow [this README](https://github.com/occlum/enable_rdfsbase/blob/master/README.md) to install `enable_rdfsbase` kernel module.
+
+
+## Install Occlum with RPM on CentOS 8.1
```
guzongmin, line 12: Update CentOS 8.1 to 8.2 to align with the latest status.

jessehui: Done. Thanks for your review.

guzongmin (approved):
LGTM

PR #205, opened 2020-11-02
Compile using sccache if it is installed

henrysun007:
Using sccache accelerates the compilation when the cache is hit. Especially, `make clean` and then `make` saves seconds.

### DIFF
--- a/src/libos/Makefile
+++ b/src/libos/Makefile
@@ -106,13 +106,19 @@ $(LIBOS_CORE_A): $(LIBOS_CORE_RS_A) $(C_OBJS) $(CXX_OBJS) $(S_OBJS) $(EDL_C_OBJS
 	@ar r $@ $(C_OBJS) $(CXX_OBJS) $(S_OBJS) $(EDL_C_OBJS)
 	@echo "GEN => $@"
 
+ifneq (, $(shell which sccache))
+    RUSTC_WRAPPER := sccache
+else
+    RUSTC_WRAPPER :=
+endif
+
 ifeq ($(OCCLUM_RELEASE_BUILD), 1)
 $(LIBOS_CORE_RS_A): $(RUST_SRCS)
-	@RUSTC_BOOTSTRAP=1 cargo build --release --target-dir=$(RUST_TARGET_DIR) -Z unstable-options --out-dir=$(RUST_OUT_DIR)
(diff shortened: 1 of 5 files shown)


tools/docker/Dockerfile.ubuntu18.04
```diff
@@ -75,7 +75,7 @@ ENV OCCLUM_RUST_VERSION=nightly-2020-09-08
 RUN curl https://sh.rustup.rs -sSf | \
         sh -s -- --default-toolchain ${OCCLUM_RUST_VERSION} -y && \
     rm -rf /root/.cargo/registry && rm -rf /root/.cargo/git && \
-    cargo -V
+    cargo install sccache
```
tatetian, line 78: I think it is good to have `cargo -V` to see the version installed.  How about keeping `cargo -V`?

src/libos/Makefile
```diff
@@ -106,13 +106,19 @@ $(LIBOS_CORE_A): $(LIBOS_CORE_RS_A) $(C_OBJS) $(CXX_OBJS) $(S_OBJS) $(EDL_C_OBJS
 	@ar r $@ $(C_OBJS) $(CXX_OBJS) $(S_OBJS) $(EDL_C_OBJS)
 	@echo "GEN => $@"
 
+RUSTC_WRAPPER :=
+
+ifneq (, $(shell which sccache))
+	RUSTC_WRAPPER=sccache
```
tatetian, line 112: There is a difference between `var := val` and `var = val` in Makefile. The style of this Makefile is to prefer the former one.

src/libos/Makefile
```diff
@@ -106,13 +106,19 @@ $(LIBOS_CORE_A): $(LIBOS_CORE_RS_A) $(C_OBJS) $(CXX_OBJS) $(S_OBJS) $(EDL_C_OBJS
...
+endif
```
tatetian, line 113: The style of this Makefile is to write this way:
```
ifneq (, $(shell which sccache))
	RUSTC_WRAPPER := sccache
else
	RUSTC_WRAPPER :=
endif
```
Let's keep the style consistent.

tatetian (commented):
Looks good. Just some minor issues. Thanks.

tatetian:
@henrysun007 Suggestion for the commit message:
```
Use sccache to accelerate Rust builds
```

tatetian (approved):
All looks good. Thanks for the contribution.

PR #206, opened 2020-11-03
Fix the conflict of symbols with glibc

henrysun007:
`tgkill` and `gettid` are defined in the high-version glibc. Refer to the man page of `tgkill` and `gettid` for more details. Besides, C macro is usually upper-case. This conflict is found in alios7u2.


tatetian (approved):
LGMT. Thanks for the contribution!

PR #208, opened 2020-11-04
Fix the alignment of stack entrypoint

liqinggd:
Libc ABI requires 16-byte alignment of the stack entrypoint

### DIFF
--- a/src/libos/src/process/do_spawn/init_stack.rs
+++ b/src/libos/src/process/do_spawn/init_stack.rs
@@ -58,10 +58,11 @@ pub fn do_init(
     let stack_buf = unsafe { StackBuf::new(stack_top, init_area_size)? };
     let envp_cloned = clone_cstrings_on_stack(&stack_buf, envp)?;
     let argv_cloned = clone_cstrings_on_stack(&stack_buf, argv)?;
+    adjust_alignment(&stack_buf, auxtbl, &envp_cloned, &argv_cloned)?;
     dump_auxtbl_on_stack(&stack_buf, auxtbl)?;
-    dump_cstrptrs_on_stack(&stack_buf, &envp_cloned);
-    dump_cstrptrs_on_stack(&stack_buf, &argv_cloned);
-    stack_buf.put(argv.len() as u64);
+    dump_cstrptrs_on_stack(&stack_buf, &envp_cloned)?;
+    dump_cstrptrs_on_stack(&stack_buf, &argv_cloned)?;
+    stack_buf.put(argv.len() as u64)?;
     Ok(stack_buf.get_pos())
 }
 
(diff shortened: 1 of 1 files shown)


src/libos/src/process/do_spawn/init_stack.rs
```diff
@@ -58,6 +58,22 @@ pub fn do_init(
...
+        let auxtbl_size = calc_auxtbl_size_on_stack(auxtbl);
+        let envp_size = calc_cstrptrs_size_on_stack(&envp_cloned);
+        let argv_size = calc_cstrptrs_size_on_stack(&argv_cloned);
+        let argc_size = mem::size_of::<u64>();
+        auxtbl_size + envp_size + argv_size + argc_size
+    };
+    // Libc ABI requires 16-byte alignment of the stack entrypoint.
+    // Current postion of the stack is 8-byte aligned already, insert 8 byte
+    // to meet the requirement if necessary.
+    if (current_pos - to_alloc_size) % 16 != 0 {
```
guzongmin, line 74: Is it better to add a function in StackBuf to adjust the alignment?

tatetian: @liqinggd The piece of code inserted here is very readable. I like it. I think it can be improved as @guzongmin points out: extract the inserted code as a separate function and give it a very readable name. It could be `adjust_alignment`. Maybe you can figure out something better.

liqinggd: Sure, I have refined the commit

src/libos/src/process/do_spawn/init_stack.rs
```diff
@@ -58,10 +58,11 @@ pub fn do_init(
     let stack_buf = unsafe { StackBuf::new(stack_top, init_area_size)? };
     let envp_cloned = clone_cstrings_on_stack(&stack_buf, envp)?;
     let argv_cloned = clone_cstrings_on_stack(&stack_buf, argv)?;
+    stack_buf.adjust_alignment(auxtbl, &envp_cloned, &argv_cloned)?;
```
tatetian, line 61: To keep the code consistent, the `adjust_alignment` should be a function (like `clone_cstrings_on_stack` and `dump_cstrptrs_on_stack `), not a method on `StackBuf`.

So why `clone_cstrings_on_stack` and `dump_cstrptrs_on_stack` were written as functions? Technically, they could also be defined as methods of `StackBuf`. The reason is thaht we want `StackBuf` to be an abstraction, which enables us to put _any_ data into a stack in a memory safe way. We can even extract `StackBuf` into a separate module so that other parts of the codebase can also use it. So `StackBuf` should not have any functionalities that are tightly bound to the specific purpose of initializing the main stack of a new process, like `dump_auxtbl_on_stack` and `dump_cstrptrs_on_stack`.

liqinggd: Make sense, I have modified this commit

tatetian (approved):

PR #210, opened 2020-11-06
Fix the return value of getcwd syscall

liqinggd:


tatetian (approved):

PR #211, opened 2020-11-06
Add the new epoll implementation

tatetian:
Before this commit, the epoll implementation works by simply delegating to the host OS through OCall. One major problem with this implementation is that it can only handle files that are backed by files on the host OS (e.g., sockets), but not those are mainly implemented by the LibOS (e.g., pipes). Therefore, a new epoll implementation that can handle all kinds of files is needed.

This commit completely rewrites the epoll implementation by leveraging the new event subsystem. Now the new epoll can handle all file types:

1. Host files, e.g., sockets, eventfd;
2. LibOS files, e.g., pipes;
3. Hybrid files, e.g., epoll files.

For a new file type to support epoll, it only needs to implement the following new methods added to [the File trait](https://github.com/tatetian/occlum/blob/dev-new_epoll/src/libos/src/fs/file.rs#L110):

* `poll` (required for all file types);
* `notifier` (required for all file types);
* `host_fd` (only required for host files);
* `recv_host_events` (only required for host files).

The core of the new epoll implementation is [`EpollFile`](https://github.com/tatetian/occlum/blob/dev-new_epoll/src/libos/src/net/io_multiplexing/epoll/epoll_file.rs#L19), whose code has been heavily documented.


tatetian:
The new epoll seems to break the gRPC tests in CI. I will look into it and find out why.

tatetian:
All CI tests passed except the gRPC test, which failed due to the limitation of SGX simulation, not this modifications introduced in this commit.

src/libos/src/net/syscalls.rs
```diff
@@ -666,23 +695,26 @@ pub fn do_epoll_wait(
...
 
     let epfile_ref = current!().file(epfd as FileDesc)?;
-    let epoll_file = epfile_ref.as_epfile()?;
-
-    let count = epoll_file.wait(&mut inner_events, timeout)?;
+    let epoll_file = epfile_ref.as_epoll_file()?;
+    let timeout = if timeout_ms >= 0 {
+        Some(Duration::from_millis(timeout_ms as u64))
+    } else {
+        None
```
guzongmin, line 712: “timeout of -1 causes epoll_wait() to block indefinitely”
Should we check the timeout?

guzongmin (approved):
LGTM

PR #215, opened 2020-11-10
Refactor Unix socket

henrysun007:
1. Implement type-safe functions;
2. Improve the correctness of nearly all the functions;
3. Improve the readability by introducing SteamServer and UnixEnd for StreamUnix.
4. Use Channel to replace the old RingBuf.


src/libos/src/net/socket/unix_socket/unix_addr.rs
```diff
@@ -0,0 +1,150 @@
...
+        let addr = unsafe { mem::MaybeUninit::<libc::sockaddr_un>::uninit().assume_init() };
+        let base = &addr as *const _ as usize;
+        let path = &addr.sun_path as *const _ as usize;
+        path - base
+    };
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum UnixAddr {
+    UnNamed,
```
tatetian, line 18: Unnamed is one word. So it should be `Unnamed`.

tatetian (commented):
These are my initial comments. I just read half of the code. Thanks.

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+impl StreamUnix {
+    pub fn new(flags: FileFlags) -> Result<Self> {
+        Ok(Self {
+            path: RwLock::new(UnixAddr::UnNamed),
+            end: SgxMutex::new(None),
+            is_blocking: AtomicBool::new(!flags.contains(FileFlags::SOCK_NONBLOCK)),
+        })
+    }
+
+    pub fn path(&self) -> UnixAddr {
```
tatetian, line 22: I think the signature of the method would be better to be like the following:
```
pub fn addr(&self) -> &UnixAddr {
```
which saves a copy and makes the name more accurate.

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+        Ok(())
+    }
+
+    //TODO: add backlog support
+    pub fn listen(&self, backlog: i32) -> Result<()> {
+        if self.path().is_unnamed() {
+            return_errno!(EINVAL, "the socket is not bound");
+        }
+
+        STREAM_SPACE.add_listener(&self.path())
```
tatetian, line 79: Why if a socket listens _twice_? Should we throw an error?

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+            return_errno!(EINVAL, "the socket is not bound");
+        }
+
+        let sock = STREAM_SPACE.pop_incoming(&self.path())?;
+
+        if flags.contains(FileFlags::SOCK_NONBLOCK) {
+            sock.set_non_blocking();
+        }
+
+        let addr = sock.peer_name().ok();
```
tatetian, line 94: There seems to be a race condition that right after the socket is accepted, its peer is closed, causing `sock.peer_name()` to fail.

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        debug!("the accepted socket is {:?} from {:?}", sock, addr);
+
+        Ok((sock, addr))
+    }
+
+    /// Backlog is temparily not supported so connect will not block
+    pub fn connect(&self, addr: &Option<UnixAddr>) -> Result<()> {
```
tatetian, line 102: Stream unix sockets's connect operation always needs an address. So the `Option` should be removed.

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+            return_errno!(EISCONN, "already connected");
+        }
+
+        let path = addr.as_ref().unwrap();
+        if *path == self.path() {
+            return_errno!(EINVAL, "self connect is not supported");
+        }
+
+        let (end_self, end_incoming) = UnixEnd::new_duplex_channel()?;
+        end_incoming.set_name(path);
```
tatetian, line 119: In this single function, one single concept has been called by three different names: addr, path, and name. This entire PR needs to be consistent about the naming.

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+        };
+
+        let socket_b = Self {
+            path: RwLock::new(UnixAddr::UnNamed),
+            end: SgxMutex::new(Some(end_b)),
+            is_blocking: AtomicBool::new(is_blocking),
+        };
+
+        Ok((socket_a, socket_b))
+    }
```
tatetian, line 171: Do we need a `shutdown` method?

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    // TODO: handle flags
+    pub fn sendto(&self, buf: &[u8], flags: SendFlags, addr: &Option<UnixAddr>) -> Result<usize> {
+        self.write(buf)
+    }
+
+    // TODO: handle flags
+    pub fn recvfrom(&self, buf: &mut [u8], flags: RecvFlags) -> Result<(usize, Option<UnixAddr>)> {
+        let data_len = self.read(buf)?;
+        let addr = self.peer_name().ok();
```
tatetian, line 147: Race condition: There is a chance that `peer_name` may not succeed even the previous `read` succeeds.

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        debug!("recvfrom {:?}", addr);
+
+        Ok((data_len, addr))
+    }
+
+    pub fn socketpair(flags: FileFlags) -> Result<(Self, Self)> {
```
tatetian, line 154: This is a constructor. Usually constructors are put in the beginning of a struct definition.

src/libos/src/net/socket/unix_socket/unix_addr.rs
```diff
@@ -0,0 +1,150 @@
+use super::*;
+use std::path::{Path, PathBuf};
+use std::{cmp, mem, slice, str};
+
+const MAX_PATH_LEN: usize = 108;
+const SUN_FAMILY_LEN: usize = mem::size_of::<libc::sa_family_t>();
+lazy_static! {
+    static ref SUN_PATH_OFFSET: usize = {
+        let addr = unsafe { mem::MaybeUninit::<libc::sockaddr_un>::uninit().assume_init() };
```
tatetian, line 9: I am not really sure that the compiler is smart enough to know that it does not need to allocate memory for `libc::sockaddr_un`.

I think it is better to use this [crate ](https://crates.io/crates/memoffset) to calculate the field offsets.

tatetian (commented):
Overall, I am positive about the overall design of this PR. Good job!

With that said, I think there are many small improvements that should be made.

src/libos/src/net/socket/unix_socket/stream/end.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    pub fn set_name(&self, name: &UnixAddr) {
+        *self.name.write().unwrap() = name.clone();
+    }
+
+    pub fn peer_name(&self) -> UnixAddr {
+        self.peer
+            .upgrade()
+            .map(|end| end.name.read().unwrap().clone())
+            .unwrap()
```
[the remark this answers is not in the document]
henrysun007: Thanks for pointing out the disconnected scenario.

src/libos/src/net/socket/unix_socket/stream/socket.rs
```diff
@@ -0,0 +1,190 @@
+use super::*;
+use alloc::sync::Arc;
+use std::fmt;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// SOCK_STREAM Unix socket
+pub struct StreamUnix {
+    path: RwLock<UnixAddr>,
+    end: SgxMutex<Option<Arc<UnixEnd>>>, // Set after connection
+    is_blocking: AtomicBool,
```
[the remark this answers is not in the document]
henrysun007: This is for when the socket is not connected.

src/libos/src/net/socket/unix_socket/stream/socket.rs
henrysun007: The value is inside a rwlock. It seems hard to borrow a value from inside rwlock.

src/libos/src/net/socket/unix_socket/stream/socket.rs
henrysun007: Basically, this is for changing the backlog. We do not implement it but it is good to support the way this API is called.

src/libos/src/net/socket/unix_socket/stream/socket.rs
henrysun007: Sure.

src/libos/src/net/socket/unix_socket/stream/socket.rs
henrysun007: Good point. It leaves inconsistency where code is changed too many times.

src/libos/src/net/socket/unix_socket/stream/socket.rs
henrysun007: `ok` converts a `Result` to an `Option`. This will panic. I do not take too much care about this kind of race condition. I do not see those test cases around, either.

src/libos/src/net/socket/unix_socket/unix_addr.rs
henrysun007: Honestly, that crate also uses `MaybeUninit`, but I used the crate to keep the code clean.

src/libos/src/net/socket/unix/stream/socket.rs
```diff
@@ -0,0 +1,207 @@
...
+        Ok(())
+    }
+
+    //TODO: add backlog support
+    pub fn listen(&self, backlog: i32) -> Result<()> {
+        if self.addr().is_unnamed() {
+            return_errno!(EINVAL, "the socket is not bound");
+        }
+
+        SERVER_SPACE.add_listener(&self.addr())
```
tatetian, line 107: What if the `listen` method is called on a connected socket? Without the aforementioned state machine, it is entirely unclear that a socket that is in a specific state behaves correctly.

src/libos/src/net/socket/unix/stream/socket.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        if self.is_connected() {
+            return_errno!(EISCONN, "already connected");
+        }
+
+        if *addr == self.addr() {
+            return_errno!(EINVAL, "self connect is not supported");
+        }
+
```
tatetian, line 144: This method is not thread-safe. If the `connect` method is called on a socket concurrently, the socket may connect to multiple different listening sockets, which does not make any sense.

src/libos/src/net/socket/unix/stream/server.rs
```diff
@@ -0,0 +1,110 @@
...
+            Addr::File(unix_path) => unix_path.absolute(),
+            Addr::Abstract(path) => addr.path_str().to_string(),
+            Addr::Unnamed => unreachable!(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Server {
+    incoming: SgxMutex<VecDeque<Stream>>,
```
tatetian, line 87: Suggestion 1. Use `Channel<Stream>` rather than `SgxMutex<VecDeque<Stream>>` so that a blocking accept and the support of backlog for listen can be implemented easily.

Suggestion 2. Logically, the incoming connections of a listening socket belongs to the socket. So I think it makes more sense to maintain the incoming connections in struct `Stream`, not a `Server` struct.

By the way, to avoid confusion, we should not introduce a type named `Server`. Technically, there are no servers here. A server should include logic to handle requests; a listening or passive socket is not a server. Since we are implementing sockets, we should use socket terms, in which "severs" are called listening or passive sockets.

 I think the  should be maintained as part of the `Stream` type, which

src/libos/src/net/socket/unix/stream/server.rs
```diff
@@ -0,0 +1,110 @@
+use super::*;
+use std::collections::btree_map::BTreeMap;
+
+#[derive(Debug)]
+pub struct Servers {
```
tatetian, line 5: I think this is better named `AddressSpace`. And `SERVER_SPACE` renamed as `ADDRESS_SPACE`.

tatetian: Second thought. `AddressRegistry` seems to be better.

henrysun007: I still prefer address space. Name space is also ok to me. `Registry` is not a word in the posix context. I have never seen the word.

src/libos/src/net/socket/mod.rs
```diff
@@ -2,18 +2,20 @@ use super::*;
 
 mod address_family;
 mod flags;
-mod host_socket;
+mod host;
 mod iovs;
 mod msg;
+mod shutdown;
```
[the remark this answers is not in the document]
henrysun007: Including shutdown makes the commit complete.

src/libos/src/net/socket/unix/stream/address_space.rs
```diff
@@ -0,0 +1,93 @@
...
+use std::collections::btree_map::BTreeMap;
+
+lazy_static! {
+    pub(super) static ref ADDRESS_SPACE: AddressSpace = AddressSpace::new();
+}
+
+pub struct AddressSpace {
+    file: SgxMutex<BTreeMap<String, Option<Arc<Listener>>>>,
+    abstr: SgxMutex<BTreeMap<String, Option<Arc<Listener>>>>,
+}
```
tatetian, line 13: As I have pointed out in my last round of comments, it is more logical for the `Listener` (previously known as `Server`), which maintains the incoming connections to a listening socket, to be part of `Stream`, instead of that of a global data structure (here, `AddressSpace`).

Admittedly, being "more logical" is not a very strong argument. So, I want to give a much stronger reason: improving performance and avoiding deadlocks. Ideally, a scalable implementation should have this property: socket A connecting to socket B should have no performance impact on socket C connecting socket D. Unfortunately, the current implementation is NOT; it has a global bottleneck--- `ADDRESS_SPACE`. Each and every connect or accept operation must hold the lock of `ADDRESS_SPACE` (see `AddressSpace::get_listener_ref`). To make it worse, the connect or accept operation may block indefinitely, which means the lock on `ADDRESS_SPACE` may be held for a very long time. Obviously, this is unacceptable.

src/libos/src/net/socket/unix/stream/stream.rs
```diff
@@ -0,0 +1,318 @@
...
+            Status::Uncon(info) => {
+                if let Some(addr) = info.addr() {
+                    ADDRESS_SPACE.add_listener(addr, cap)?;
+                    *inner = Status::Lis(addr.clone());
+                } else {
+                    return_errno!(EINVAL, "the socket is not bound");
+                }
+            }
+            Status::Con(_) => return_errno!(EINVAL, "the socket is already connected"),
+            /// Modify the capability of the channel holding incoming sockets
```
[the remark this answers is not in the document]
henrysun007: It is a test case of gvisor that passed on linux. The source code is [here](https://github.com/google/gvisor/blob/cc522a9cfb0a2d5976a3a592cba8e9d23b06c6ce/test/syscalls/linux/accept_bind.cc#L101).

src/libos/src/net/socket/unix/addr.rs
```diff
@@ -0,0 +1,133 @@
+use super::*;
+use std::path::{Path, PathBuf};
+use std::{cmp, mem, slice, str};
+
+const MAX_PATH_LEN: usize = 108;
+const SUN_FAMILY_LEN: usize = mem::size_of::<libc::sa_family_t>();
+lazy_static! {
+    static ref SUN_PATH_OFFSET: usize = memoffset::offset_of!(libc::sockaddr_un, sun_path);
+}
```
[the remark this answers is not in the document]
henrysun007: Basically, offset and size are different. They were same only if there was no padding after `sa_family_t`.

src/libos/src/net/socket/unix/addr.rs
```diff
@@ -0,0 +1,133 @@
...
+        let mut addr: libc::sockaddr_un = unsafe { mem::zeroed() };
+        addr.sun_family = AddressFamily::LOCAL as libc::sa_family_t;
+
+        let addr_len = match self {
+            Self::File(unix_path) => {
+                let path_str = unix_path.path_str();
+                let buf_len = path_str.len();
+                addr.sun_path[..buf_len]
+                    .copy_from_slice(unsafe { &*(path_str.as_bytes() as *const _ as *const [i8]) });
+                buf_len + *SUN_PATH_OFFSET + 1
```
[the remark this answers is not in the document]
henrysun007: I will add some comments to the code.

src/libos/src/net/socket/unix/stream/address_space.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn get_space(&self, addr: &Addr) -> SgxMutexGuard<'_, BTreeMap<String, Option<Arc<Listener>>>> {
+        match addr {
+            Addr::File(unix_path) => self.file.lock().unwrap(),
+            Addr::Abstract(path) => self.abstr.lock().unwrap(),
+        }
+    }
+
+    fn get_key(addr: &Addr) -> String {
+        match addr {
+            Addr::File(unix_path) => unix_path.absolute(),
```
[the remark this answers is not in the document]
henrysun007: Thanks for pointing this out.

src/libos/src/net/socket/unix/addr.rs
```diff
@@ -0,0 +1,133 @@
...
+const MAX_PATH_LEN: usize = 108;
+const SUN_FAMILY_LEN: usize = mem::size_of::<libc::sa_family_t>();
+lazy_static! {
+    static ref SUN_PATH_OFFSET: usize = memoffset::offset_of!(libc::sockaddr_un, sun_path);
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Addr {
+    File(UnixPath),
+    Abstract(String),
```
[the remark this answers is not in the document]
henrysun007: `Vec` makes the code unclear. If the length was larger than what the user means, say 108, there would be many redundant paddings.  It is not that type clear. I will add some comments to express this principle.

src/libos/src/net/socket/unix/stream/stream.rs
```diff
@@ -0,0 +1,318 @@
...
+            }
+            Status::Con(endpoint) => return_errno!(EISCONN, "already connected"),
+            Status::Lis(addr) => return_errno!(EINVAL, "invalid socket for connect"),
+        }
+    }
+
+    pub fn accept(&self, flags: FileFlags) -> Result<(Self, Option<Addr>)> {
+        match &*self.inner() {
+            Status::Lis(addr) => {
+                let endpoint = ADDRESS_SPACE.pop_incoming(&addr)?;
```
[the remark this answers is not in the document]
henrysun007: I will fix the blocking issues with `poll` support in the near future. I left some comments in this commit.

src/libos/src/net/socket/unix/addr.rs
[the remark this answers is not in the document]
tatetian: Good point. Let's stick with your way.

tatetian (approved):
Good enough to be accepted. Needs more commits to fully implement unix sockets.

PR #217, opened 2020-11-11
Add the new poll implementation

tatetian:
Based on the new event subsystem, the new poll implementation can work with any file types, including LibOS files and host files.

The new poll syscall can be found [here](https://github.com/occlum/occlum/blob/ea806f33690abe986b760672b56cf2b44f9125da/src/libos/src/net/io_multiplexing/poll_new/mod.rs#L13), which is as simple as 100 lines. The is because the heavy-lifting is done by a helper type named [`EventMonitor`](https://github.com/occlum/occlum/blob/ea806f33690abe986b760672b56cf2b44f9125da/src/libos/src/net/io_multiplexing/poll_new/event_monitor.rs#L16), whose code is commented intensively.

The old poll implementation has not been removed (yet) as some code still relies on it, for example, the unix sockets. The unix sockets will be soon rewritten based the primitives provided by the new event subsystem. So the poll test for unix sockets are disabled in this commit and they will be enabled when the unix socket writes are done.

### DIFF
--- a/src/libos/src/fs/dev_fs/dev_null.rs
+++ b/src/libos/src/fs/dev_fs/dev_null.rs
@@ -16,6 +16,10 @@ impl File for DevNull {
         Ok(bufs.iter().map(|buf| buf.len()).sum())
     }
 
+    fn poll_new(&self) -> IoEvents {
+        IoEvents::IN
+    }
+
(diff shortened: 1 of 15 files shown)


guzongmin (approved):
LGTM

PR #218, opened 2020-11-11
Fix two cpuid bugs

henrysun007:
1.  `>>` has higher precedence than `&`. Use parentheses to conduct `&` first;
2. In the latest Intel software developer's manual, cpuid leaf 06H EDX
is related to the logical processor.


tatetian (approved):

PR #222, opened 2020-11-14
Rewrite the select syscall using the new poll implementation

tatetian:
In addition to the new select implementation, this PR also contains two bugfixes for `Channel`.

### DIFF
--- a/src/libos/src/fs/events.rs
+++ b/src/libos/src/fs/events.rs
@@ -1,3 +1,5 @@
+use atomic::{Atomic, Ordering};
+
 use crate::events::{Event, EventFilter, Notifier, Observer};
 use crate::prelude::*;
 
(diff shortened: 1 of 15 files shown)


src/libos/src/fs/events.rs
```diff
@@ -38,6 +40,32 @@ impl IoEvents {
...
 impl Event for IoEvents {}
 
+pub trait AtomicIoEvents {
+    /// Update the IoEvents atomically.
+    ///
+    /// The update is equivalent to the following assignment
+    /// ```
+    /// self.store(self.load(Ordering::Relaxed) & !*mask | *ready, ordering)
+    /// ```
+    fn update(&self, ready: &IoEvents, mask: &IoEvents, ordering: Ordering);
```
guzongmin, line 50: Do not get the point why do we need to give user the flexibility to change the ordering?

guzongmin (approved):
LGTM

PR #224, opened 2020-11-16
Init the memory for elf to zero

liqinggd:


tatetian (approved):

PR #226, opened 2020-11-16
Add getdents syscall

liqinggd:


src/libos/src/fs/file_ops/dirent.rs
```diff
@@ -1,19 +1,99 @@
...
-    type_: u8,
-    /// Filename (null-terminated)
-    name: [u8; 0],
+    type_: T,
+}
+
+impl<T: DirentType + Copy + Default> LinuxDirent<T> {
+    fn new(ino: u64, name: &str) -> Self {
+        let ori_len = ::core::mem::size_of::<LinuxDirent<T>>() + name.len() + 1;
+        let len = (ori_len + 7) / 8 * 8; // align up to 8 bytes
```
tatetian, line 65: Actually, we have an [`align_up`](https://github.com/occlum/occlum/blob/55eaae88101b5760cdd3545e82824afd84ffd0ec/src/libos/src/prelude.rs#L34) function already.

src/libos/src/fs/file_ops/dirent.rs
```diff
@@ -1,19 +1,99 @@
...
     /// Inode number
     ino: u64,
     /// Offset to next structure
     offset: u64,
     /// Size of this dirent
     reclen: u16,
     /// File type
-    type_: u8,
-    /// Filename (null-terminated)
-    name: [u8; 0],
```
tatetian, line 14: I suggest keeping the `name` field, although it is zero length and thus "useless" in practice. But it can serve as a document to remind us that `LinuxDirent` actually has such a field.

liqinggd: Ok

src/libos/src/fs/file_ops/dirent.rs
liqinggd: Thanks for your suggestion, I find this helper function in `prelude.rs`

tatetian (approved):

PR #227, opened 2020-11-16
Enable optimized string and math libs

guzongmin:


tatetian (approved):

PR #232, opened 2020-11-18
Add SGX hardware mode test

jessehui:
Currently, only a subset of tests are enabled. Other tests will be enabled in the future.

Results can be checked here: https://github.com/jessehui/occlum/actions/runs/369797484


guzongmin (approved):
LGTM

PR #246, opened 2020-11-23
Change Tcmalloc as optional feature

guzongmin:
The current Tcmalloc has memory leak issue. So change it as optional. By
default, dlmalloc is used. Enable tcmalloc with below command:
make TCMALLOC=Y


tatetian (approved):
LGTM

PR #248, opened 2020-11-30
Add support to mount FS at any existing directories

### LINKED ISSUE #161: The mount config cannot support mounting FS at a path which is not a child of the root.
Currently,  the mount config only support to mount FS at `"/dir_name" `, the `dir_name` must be a child of the root.

We should support to mount FS at `"/dir_name/child_dir"`, there should be no restriction here, if the dir exists in the image, it should be OK to mount the FS.

liqinggd:
Fixes #161


guzongmin (approved):
LGTM

PR #250, opened 2020-12-01
Optimize the loading process of the ELF file

liqinggd:
1. Load ld.so according to the executable automatically
2. Add the position-independent check for ELF file


liqinggd:
This PR is already included in https://github.com/occlum/occlum/pull/263, close this one

Issue #244, opened 2020-11-23
[BUG]FISH demo test hang


jessehui:
# Describe the bug
When using Occlum 0.18.0 to test FISH demo, the process hangs.

# To reproduce
1. Follow the readme of FISH demo to download and install all dependencies
2. execute `./run_fish_test.sh` and after occlum is running, nothing is printed out and it will hang
3. Change the first line of `fish_script.sh` to `command echo "Hello-world-from-fish" > /root/output.txt` and everything is normal
4. 

# Expected behavior
I expect running `./run_fish_test.sh` will print `Hello world from fish`.

# Logs
The last syscall log before hanging is about read:
```
[2020-11-23T09:15:40.721Z][TRACE][T3][#13][····Read] Syscall { num = Read, fd = 0, buf = 0x7fc59aa96ca0, size = 463 }
[2020-11-23T09:15:40.721Z][DEBUG][T3][#13][····Read] read: fd: 0
```
And after this, no new log shows.

# Environment

- HW: [e.g. SGX1, SGX2] SGX1
- OS: [e.g. Ubuntu18.04, CentOS8.1] Ubuntu18.04
- Occlum version: [e.g. 0.17.0] 0.18.0

# Additional context

# Possible solution/Implementation
Maybe pipe is not fully functional?

tatetian:
Thanks for reporting the problem. I will try to reproduce the problem and find out the root cause.

tatetian:
This bug is fixed in #254 .

guzongmin:
fixed in 0.18.1

PR #253, opened 2020-12-01
Fix a bug of channels

tatetian:
This is a bugfix for #244 . It ensures that when an object of Producer/Consumer for channels is dropped, its shutdown method is called automatically. This ensures that the peer of a Producer/Consumer gets notified and won't wait indefinitely.


jessehui:
Verified. This patch can fix https://github.com/occlum/occlum/issues/244. Thank you!

guzongmin (approved):

PR #255, opened 2020-12-02
Fix memory leak

guzongmin:


tatetian (approved):

PR #263, opened 2020-12-04
Enable Glibc support

### LINKED ISSUE #214: Support Glibc
Currently, Only musl-based applications can run on Occlum, many Linux distributions use Glibc as the C Library. Applications must be recompiled with `occlum-gcc` to use Occlum. To overcome this inconvenience, we should support Glibc.

liqinggd:
This PR enables Glibc as an optional libc for Apps running on Occlum. 

Fixes #214 

Please check out the PR in https://github.com/occlum/glibc/pull/1 for Glibc's modification.


tools/toolchains/glibc/build.sh
```diff
@@ -0,0 +1,26 @@
...
+mkdir -p ${SRC_DIR}
+cd ${SRC_DIR}
+# Download glibc
+git clone -b occlum-glibc-2.27 https://github.com/occlum/glibc .
+
+mkdir -p ${BUILD_DIR}
+cd ${BUILD_DIR}
+# Build and install glibc
+unset LD_LIBRARY_PATH
+CFLAGS="-O2 -Wp,-U_FORTIFY_SOURCE -fno-stack-protector -Wno-unused-value -g" ${SRC_DIR}/configure \
```
guzongmin, line 23: Is it possible to enable stack-protector and enable tests?

liqinggd: Add new commit ”Initialize a randon number on stack for the stack protector“ to support stack protector

src/libos/src/util/random.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub fn get_random(rand: &mut [u8]) -> Result<()> {
+    extern "C" {
+        fn sgx_read_rand(rand_buf: *mut u8, buf_size: usize) -> sgx_status_t;
+    }
+
+    let status = unsafe { sgx_read_rand(rand.as_mut_ptr(), rand.len()) };
+    match status {
+        sgx_status_t::SGX_SUCCESS => Ok(()),
+        _ => return_errno!(EAGAIN, "failed to get random number from SGX"),
```
guzongmin, line 13: If you return EAGAIN, then Occlum would crash.

liqinggd: I just move the code from `dev_random.rs` to random.rs, why Occlum will crash?

guzongmin (dismissed):
LGTM

yuanwu2017:
I have tested the flink with the glibc. But the flink cannot work.  It is caused by the gethostbyname error. I updated a test sample in my github. Please have a try. 
 https://github.com/yuanwu2017/occlum/tree/benchmark/demos/benchmark/test_host

liqinggd:
> I have tested the flink with the glibc. But the flink cannot work. It is caused by the gethostbyname error. I updated a test sample in my github. Please have a try.
> https://github.com/yuanwu2017/occlum/tree/benchmark/demos/benchmark/test_host

I @yuanwu2017 , I have reproduced your problem with env `OCCLUM_LOG_LEVEL=trace`, the log shows that it lacks `libnss_files.so.2`, please add `cp $occlum_glibc/libnss_files.so.2 image/$occlum_glibc` in the bash script.

yuanwu2017:
> > I have tested the flink with the glibc. But the flink cannot work. It is caused by the gethostbyname error. I updated a test sample in my github. Please have a try.
> > https://github.com/yuanwu2017/occlum/tree/benchmark/demos/benchmark/test_host
> 
> I @yuanwu2017 , I have reproduced your problem with env `OCCLUM_LOG_LEVEL=trace`, the log shows that it lacks `libnss_files.so.2`, please add `cp $occlum_glibc/libnss_files.so.2 image/$occlum_glibc` in the bash script.

Thanks Qing, You are right. The flink works.

src/libos/src/process/syscalls.rs
```diff
@@ -85,6 +86,154 @@ fn clone_file_actions_safely(fdop_ptr: *const FdOp) -> Result<Vec<FileAction>> {
...
+}
+
+#[repr(C)]
+struct SpawnAction {
+    tag: u32,
+    action: Action,
+}
+
+impl SpawnAction {
+    // Wrappers to access the union field safely
```
tatetian, line 129: I admire the good intention of defining these wrapper functions. But I am afraid that they are not as useful as they are intended to be.

First, they make the code unnecessarily complex. And defining each of the wrapper functions takes four lines of code, but they are all essentially one-liner and they are only used once. So defining the wrapper functions does not deliver any benefit  of code reuse.

Second, the `assert` statements are pointless. Because each wrapper is only used once and used in the correct branches of `match sa.tag`. So they cannot deliver any real benefit of safety.

I recommend to rewrite as followinng:

```rust
#[repr(C)]
struct SpawnAction {
    tag: u32,
    action: Action,
}

impl SpawnAction {
    pub fn to_file_action(&self) -> Result<FileAction> {
        match self.tag {
            SPAWN_DO_CLOSE => FileAction::Close(unsafe { self.action.close_action.fd }),
            SPAWN_DO_DUP2 => FileAction::Dup2(unsafe { ... }),
            SPAWN_DO_OPEN => FileAction::Open {
                path: clone_cstring_safely(unsafe { ... })?
                    .to_string_lossy()
                    .into_owned(),
                mode: unsafe { ... },
                oflag: unsafe { ... },
                fd: unsafe { ... },
            },
            _ => return_errno!(EINVAL, "Unknown file action tag"),
        };
    }
}
```

The above version removes all (useless) wrappers without sacrificing memory safety. And `clone_file_actions_from_fa_safely` looks more neat.

Makefile
```diff
@@ -54,7 +54,10 @@ src:
 	@$(MAKE) --no-print-directory -C src
 
 test:
-	@$(MAKE) --no-print-directory -C test test
+	@CC=occlum-gcc CXX=occlum-g++ $(MAKE) --no-print-directory -C test test
+
+test-glibc:
+	@CC=gcc CXX=g++ $(MAKE) --no-print-directory -C test test
```
tatetian, line 60:
```
$ cd <occlum_project> && make test
```
should be equivalent to
```
$ cd <occlum_project>/test && make test
```
That is, the root Makefile simply redirects commands to sub Makefiles. So the Glibc support should apply equally well to both the root Makefile and the sub Makefile in `test/`.

One way to achieve this is to add an env var named `OCCLUM_TEST_GLIBC`. Use this env var to decide the values of `CC` and `CXX` in `test/Makefile`.

tools/docker/Dockerfile.centos8.1
```diff
@@ -83,7 +90,7 @@ COPY toolchains/rust /tmp/rust
 WORKDIR /tmp
 RUN cd rust && ./build.sh && rm -rf /tmp/rust
 ENV PATH="/opt/occlum/toolchains/rust/bin:$PATH"
-ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/opt/intel/sgxsdk/sdk_libs"
```
tatetian, line 86: Why replacing the path instead of inheriting?

tools/docker/Dockerfile.ubuntu18.04
```diff
@@ -84,6 +86,11 @@ WORKDIR /tmp
 RUN cd gcc && ./build.sh && ./install_zlib.sh && rm -rf /tmp/gcc
 ENV PATH="/opt/occlum/build/bin:/usr/local/occlum/bin:$PATH"
 
+# Install glibc
+COPY toolchains/glibc /tmp/glibc
+WORKDIR /tmp
+RUN cd glibc && ./build.sh && rm -rf /tmp/glibc
```
tatetian, line 92: According to your proposal, the toolchain folder would look like this
```
clang/
gcc/
glibc/
golang/
java/
rust/
```
You see the issue here, right? `clang` and `gcc` are different C compilers; so they are exclusive. But `glibc` are libc. Adding a subfolder named `glibc` would immediately raise a question: what kinds of libc does `clang` or `gcc` use or they do not include a libc at all? To avoid any such confusion, I think we should rename `gcc/` to `musl-gcc/` and you can keep the name of the new folder as `glibc/` (which is a very precise description as it does contain GCC, only the libc). Furthermore, we can later remove the `clang` folder, which has been deprecated anyway.

tools/occlum
```diff
@@ -129,20 +129,43 @@ cmd_init() {
...
+    cp -t image/lib \
         /lib/ld-musl-x86_64.so.1 \
-        "$occlum_gcc_lib/libc.so" \
-        "$occlum_gcc_lib/libstdc++.so.6" \
-        "$occlum_gcc_lib/libgcc_s.so.1" \
-        "$occlum_gcc_lib/libgomp.so.1"
+        "$occlum_musl_lib/libc.so" \
+        "$occlum_musl_lib/libstdc++.so.6" \
+        "$occlum_musl_lib/libgcc_s.so.1" \
+        "$occlum_musl_lib/libgomp.so.1"
```
tatetian, line 153: Since we demand copying more and more files, maybe we should consider using symbolic links instead.

tools/toolchains/glibc/build.sh
```diff
@@ -0,0 +1,26 @@
...
+cd ${SRC_DIR}
+# Download glibc
+git clone -b occlum-glibc-2.27 https://github.com/occlum/glibc .
+
+mkdir -p ${BUILD_DIR}
+cd ${BUILD_DIR}
+# Build and install glibc
+unset LD_LIBRARY_PATH
+CFLAGS="-O2 -g" ${SRC_DIR}/configure \
+--prefix=${INSTALL_DIR} --with-tls --without-selinux --enable-stack-protector=strong --disable-nscd
```
tatetian, line 24: Adding an indent would be better.

tools/occlum
```diff
@@ -129,20 +129,43 @@ cmd_init() {
     mkdir -p image
     mkdir -p image/bin
     mkdir -p image/lib
+    mkdir -p image/lib64
     mkdir -p image/root
     mkdir -p image/host
     mkdir -p image/tmp
+    local occlum_glibc_lib=/usr/local/occlum-glibc/lib
```
tatetian, line 136: I am surprised that this is not under `/usr/local/occlum` and that we are having a new folder under `/usr/local/`.

The original reason of installing the gcc-musl toolchain at `/usr/local` is to follow the convention of the installation path for toolchains. And also that the headers and libraries of the toolchain should be visible to users. In contrast, all contents in `/opt/occlum` is not supposed to be accessed directly by users.

I think the new glibc is more akin to the contents in `/opt/occlum`, which are internal to Occlum. So the new glibc should be installed in `/opt/occlum`. Make sense?

tatetian (commented):
The commit looks good in general. I found some minor issues.

src/libos/src/process/do_spawn/init_stack.rs
```diff
@@ -53,11 +54,13 @@ pub fn do_init(
...
     argv: &[CString],
     envp: &[CString],
-    auxtbl: &AuxVec,
+    auxtbl: &mut AuxVec,
 ) -> Result<usize> {
     let stack_buf = unsafe { StackBuf::new(stack_top, init_area_size)? };
     let envp_cloned = clone_cstrings_on_stack(&stack_buf, envp)?;
     let argv_cloned = clone_cstrings_on_stack(&stack_buf, argv)?;
+    let rand_val_ptr = generate_random_on_stack(&stack_buf)?;
+    auxtbl.set(AuxKey::AT_RANDOM, rand_val_ptr as *const () as u64);
```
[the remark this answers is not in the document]
liqinggd: This is because the AuxKey::AT_RANDOM in aux table is a pointer that points to a value in stack. The pointer must be initialized on the stack before setting the aux table.

To make the aux_tbl immutable, maybe I can change the stack from one step to two steps: 
init_stack_phase1() --> aux_table::init(random_value_ptr_on_stack) --> init_stack_phase2()

Is it make sense?

src/libos/src/process/syscalls.rs
liqinggd: The reason why I choose to add wrapper functions is as follows:
The wrapper functions can ensure the unsafe value can be accessed safely, if users access the `SpawnAction` by mistake, it can panic.

The `to_file_action` in your suggestion looks simpler, good idea

tools/docker/Dockerfile.centos8.1
liqinggd: This makes the `LD_LIBRARY_PATH =:/opt/intel/sgxsdk/sdk_libs`, the first element is `(null)`.
If I installed <glibc-for-occlum-dir> in Docker container, then I `cd` into the dir, any bash commands will fail because the commands will use the ld.so or libs in the <glibc-for-occlum-dir>, which is the current directory.

src/libos/src/util/random.rs
```diff
@@ -0,0 +1,27 @@
...
+    // sgx_read_rand() may fail because of HW failure of RDRAND instruction,
+    // add tries to get the random number.
+    for _ in 0..MAX_RETRIES {
+        let status = unsafe { sgx_read_rand(rand.as_mut_ptr(), rand.len()) };
+        match status {
+            sgx_status_t::SGX_SUCCESS => {
+                return Ok(());
+            }
+            sgx_status_t::SGX_ERROR_INVALID_PARAMETER => {
+                return_errno!(EINVAL, "failed to get random number from SGX");
```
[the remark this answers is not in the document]
liqinggd: Yes, panic here makes sense, I should also modify the `read` in dev_random.rs to check the buffer length before using it, if the length is zero, just return zero.

src/libos/src/process/elf_file.rs
```diff
@@ -45,18 +49,29 @@ impl<'a> ElfFile<'a> {
     }
 }
 
-pub trait ProgramHeaderExt {
+pub trait ProgramHeaderExt<'a> {
     fn loadable(&self) -> bool;
+    fn is_interpreter(&self) -> bool;
     fn validate(&self) -> Result<()>;
+    fn get_content(&self, elf_file: &ElfFile<'a>) -> SegmentData<'a>;
 }
```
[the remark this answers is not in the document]
liqinggd: I can try this

src/libos/src/process/syscalls.rs
```diff
@@ -9,6 +9,7 @@ use crate::time::{timespec_t, ClockID};
 use crate::util::mem_util::from_user::*;
 use std::ptr::NonNull;
 
+// Syscall handler for musl's spawn
```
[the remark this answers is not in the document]
liqinggd: It is Ok

src/libos/src/process/do_spawn/init_stack.rs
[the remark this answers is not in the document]
tatetian: I see. Then I suggest keeping the current code.

src/libos/src/util/random.rs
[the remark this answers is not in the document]
tatetian: Agree.

tatetian (approved):
LGTM.

tools/occlum
liqinggd: Right

tools/toolchains/glibc/build.sh
liqinggd: OK

tools/docker/Dockerfile.ubuntu18.04
liqinggd: OK

PR #266, opened 2020-12-11
Fix the UDS path is too long issue

guzongmin:
The UDS path length should less than sun_path length, it
is too short in some cases. So use a file under /tmp folder
as the UDS path.

### DIFF
--- a/src/exec/src/bin/occlum_exec_client.rs
+++ b/src/exec/src/bin/occlum_exec_client.rs
@@ -18,9 +18,7 @@ use occlum_exec::occlum_exec::{
     KillProcessRequest, StopRequest,
 };
 use occlum_exec::occlum_exec_grpc::OcclumExecClient;
-use occlum_exec::{
-    DEFAULT_CLIENT_FILE, DEFAULT_SERVER_FILE, DEFAULT_SERVER_TIMER, DEFAULT_SOCK_FILE,
-};
+use occlum_exec::{DEFAULT_SERVER_FILE, DEFAULT_SERVER_TIMER, DEFAULT_SOCK_FILE};
 use protobuf::RepeatedField;
 use sendfd::SendWithFd;
 use signal_hook::iterator::Signals;
@@ -142,7 +140,12 @@ fn start_server(client: &OcclumExecClient, server_name: &str) -> Result<u32, Str
             Err(_resp) => {
                 if !server_launched {
                     debug!("server is not running, try to launch the server.");
-                    match Command::new(server_name).stdout(Stdio::null()).spawn() {
+                    match Command::new(server_name)
+                        .arg("-d")
+                        .arg(env::current_dir().unwrap())
+                        .stdout(Stdio::null())
+                        .spawn()
+                    {
                         Err(_r) => {
                             return Err("Failed to launch server".to_string());
                         }
@@ -233,6 +236,14 @@ fn main() -> Result<(), i32> {
 
     let matches = App::new("Occlum")
         .version("0.1.0")
+        .arg(
+            Arg::with_name("instance_dir")
+                .short("d")
+                .long("instance_dir")
+                .takes_value(true)
+                .default_value("./")
+                .help("The Occlum instance dir."),
+        )
         .subcommand(
             App::new("start").about(
                 "Start the Occlum server. If the server already running, immediately return.",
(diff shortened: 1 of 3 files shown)


src/exec/src/bin/occlum_exec_client.rs
```diff
@@ -269,14 +271,21 @@ fn main() -> Result<(), i32> {
...
+        sockfile.as_mut_str(),
         DEFAULT_CLIENT_FILE,
         DEFAULT_SOCK_FILE,
     );
 
-    let client = OcclumExecClient::new_plain_unix(&sock_file, ClientConf::new())
+    let mut hasher = DefaultHasher::new();
+    sockfile.hash(&mut hasher);
+
+    let mut sockfile = String::from("/tmp/occlum_");
```
kailun-qin, line 284: `/tmp` may not be a good place for UDS files since OS/admins may remove any file within it at any moment. Can we use e.g. `/run` instead?
See FHS suggestions [/run](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html) and [/tmp](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s18.html) for further details.

src/exec/src/bin/occlum_exec_server.rs
```diff
@@ -61,8 +63,15 @@ fn main() {
         DEFAULT_SOCK_FILE,
     );
 
+    let mut hasher = DefaultHasher::new();
+    sockfile.hash(&mut hasher);
+
+    let mut sockfile = String::from("/tmp/occlum_");
+    sockfile.push_str(hasher.finish().to_string().as_str());
+    let sockfile = &*sockfile;
```
kailun-qin, line 71: String has already implemented `Deref<Target=str>`. I guess we do not necessarily need this here and can directly use check_server_status(&sockfile) instead.
Same above for the occlum_exec_client.rs.

src/exec/src/bin/occlum_exec_client.rs
guzongmin: Thanks. Change the path.

src/exec/src/bin/occlum_exec_server.rs
guzongmin: line 71 is used to change sockfile to be immutable. Just make it more clear to myself.

tools/occlum
```diff
@@ -248,6 +248,10 @@ cmd_start() {
     if [ "`get_enclave_debuggable_flag`" == "false" ]; then
         export OCCLUM_RELEASE_ENCLAVE=1
     fi
+
+    # /run/occlum is needed to save the occlum socket files for exec
+    mkdir -p /run/occlum
```
tatetian, line 253: You sure a normal, non-root user (without sudo) can create directories under `/run`?

src/exec/src/bin/occlum_exec_server.rs
tatetian: To be honest, I am quite confused with `sockfile`: four variables named `sockfile` is defined at line 59, 60, 69, and 71, respectively. The content and meaning of sockfile keeps changing (rapidly), which is a bad sign for readability and correctness. I think they really should have different names to make their differences and relationship more clear and straightforward. The client code has the same issue.

src/exec/src/bin/occlum_exec_client.rs
```diff
@@ -269,14 +271,21 @@ fn main() -> Result<(), i32> {
         .map(|(key, val)| format!("{}={}", key, val))
         .collect();
 
-    let mut sock_file = String::from(args[0].as_str());
-    let sock_file = str::replace(
-        sock_file.as_mut_str(),
+    let mut sockfile = String::from(args[0].as_str());
```
tatetian, line 274: What is the meaning of `args[0]`? It seems that a pair of client and server needs to have the same value of `args[0]` to generate the same hash.

src/exec/src/bin/occlum_exec_server.rs
guzongmin: updated

tools/occlum
guzongmin: Seems that "/tmp" is still the best choice.

src/exec/src/bin/occlum_exec_client.rs
guzongmin: updated

guzongmin: Change the solution. Add a parameter "instance_dir" for both client and server.

tatetian (approved):

PR #268, opened 2020-12-13
Rewrite exit/wait4 and futex using the event subsystem

tatetian:
The event subsystem is intended to serve as an easy-to-use and reliable foundation for event notification. Currently, most event-handling related code (e.g., sigtimedwait, pipe, poll, epoll, etc.) has been implemented using the event subsystem; but there are two exceptions: exit/wait4 and futex.

This PR rewrites exit/wait4 and futex using the event subsystem. The direct benefit of this PR is making the code simpler (the lines of code deleted is more than those added) and more robust (at least one bug in futex is fixed). More importantly, this PR paves the way for  enabling async/.await. With all code that involves event handling or wait/wakeup now based on the event subsystem, our task of making event handling async become easier.


src/libos/src/process/syscalls.rs
```diff
@@ -142,20 +144,43 @@ pub fn do_futex(
...
+        let is_absolute = if futex_op != FutexOp::FUTEX_WAIT {
+            true
         } else {
-            ClockID::CLOCK_MONOTONIC
+            false
         };
-        Ok(Some(FutexTimeout::new(clock_id, ts)))
+        if is_absolute {
+            // TODO: use a secure clock to transfer the real time to monotonic time
+            let clock_id = if futex_flags.contains(FutexFlags::FUTEX_CLOCK_REALTIME) {
```
liqinggd, line 167: Actually, I have implemented a commit using just one timeout, see [here](https://github.com/liqinggd/occlum/commit/451f2047774f5b665204b61ebe8ce87f8edfbc28) for the detail.
The reason why I choose to add the `clockbit` argument is that currently, we do not have a secure clock or vdso, so the `clock_get_time` Ocall is too slow for `futex`, so I add the TODO here.

src/libos/src/process/syscalls.rs
```diff
@@ -142,20 +144,43 @@ pub fn do_futex(
...
+                ClockID::CLOCK_REALTIME
+            } else {
+                ClockID::CLOCK_MONOTONIC
+            };
+            let now = crate::time::do_clock_gettime(clock_id)
+                .unwrap()
+                .as_duration();
+            timeout = timeout
+                .checked_sub(now)
+                .ok_or_else(|| errno!(EINVAL, "timeout is invalid"))?;
```
liqinggd, line 177: I think we should return ETIMEDOUT here because `timeout` is already validated here. If the `checked_sub` fails, this means the timeout is expired.

src/libos/src/events/waiter_queue.rs
```diff
@@ -13,15 +13,15 @@ use crate::prelude::*;
 /// of `Waker`s.
 pub struct WaiterQueue {
     count: AtomicUsize,
-    wakers: SgxMutex<VecDeque<Waker>>,
+    wakers: SgxMutex<LinkedList<Waker>>,
```
kailun-qin, line 16: Would you please kindly elaborate more on why we'd prefer `LinkedList` to `VecDeque` for `wakers`? On which case will it better perform?

PR #275, opened 2020-12-16
Change SGX hardware mode CI test trigger type

jessehui:
Currently, SGX hardware CI tests only run when a new commit is pushed to an `occlum/occlum` 's branch. This commit expands the trigger type to pull request with labeled `SGX-hardware-test-required`. Since only the owners have the privilege to label the pull request, this won't make the hardware test insecure.


guzongmin (approved):
LGTM

PR #276, opened 2020-12-16
Add support for DCAP

henrysun007:
1. Five new ioctl commands of /dev/sgx are provided for occlum
applications to securely get and verify DCAP quote;
2. Not all the functions of the intel DCAP package are open to
developers to simplify the DCAP usage;
3. The test may only run on the platform with DCAP driver installed;
4. A macro OCCLUM_DISABLE_DCAP is used to separate the DCAP code from
the other code.

### DIFF
--- a/src/Enclave.edl
+++ b/src/Enclave.edl
@@ -10,6 +10,28 @@ enclave {
     include "sgx_quote.h"
     include "occlum_edl_types.h"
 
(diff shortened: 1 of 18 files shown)


guzongmin (approved):
LGTM

The code needs to pass the CI before check in the master. So would create a branch to test it.

PR #277, opened 2020-12-16
TEST for label

jessehui:


jessehui:
Looks like tag trigger only works for new pull requests.

PR #284, opened 2020-12-18
Fix the sched test when available cpu is less than online cpu

liqinggd:
The number of available cpu is fetched from `sched_getaffinity()`, it may be less than all online cpus, we should not assume that the number is equal to all online cpus.


tatetian (approved):
LGTM

PR #297, opened 2020-12-30
Fix multiplication overflow

henrysun007:
The added check can detect negative count, too.


guzongmin:
I do not find any case failed when you add this check, why do we need it?

henrysun007:
The additional check make the code robust, no more overflow will happen. Another benefit is it will return `EINVAL` when `count = -1`. I encounter the overflow panic with gvisor test [here](https://github.com/google/gvisor/blob/fc153750ed8b59e9fede94d17102346992396ecb/test/syscalls/linux/poll.cc#L47) but not limited to this scenario.

PR #306, opened 2021-01-08
Enable gvisor tests

guzongmin:


src/libos/src/net/socket/unix/stream/endpoint.rs
```diff
@@ -100,6 +103,38 @@ impl Inner {
...
 
+    pub fn poll(&self) -> IoEvents {
+        let mut events = IoEvents::empty();
+
+        /// get the status of peer
+        if self.reader.is_peer_shutdown() {
+            events |= IoEvents::RDHUP;
+            if self.writer.is_peer_shutdown() {
+                events |= IoEvents::HUP;
+            }
```
tatetian, line 114: Why the  `HUP` case is embedded inside the `RDHUP` case?

src/libos/src/net/socket/unix/stream/endpoint.rs
```diff
@@ -100,6 +103,38 @@ impl Inner {
...
+        }
+
+        let reader_events = self.reader.poll();
+        let writer_events = self.writer.poll();
+        events |= (reader_events & IoEvents::IN) | (writer_events & IoEvents::OUT);
```
tatetian, line 119: What exactly is the relationship between the events of an endpoint and its underlying reader and writer? If some kind of conversion or filter should be done, then `EndpointObserver::on_event` should also reflect this. Since we have full control over the underlying `Channel`, it is possible for us to make the events on `Channel` consistent with the events  on `Endpoint`.

PR #311, opened 2021-01-14
Use pthread_join to wait for the async thread to exit

henrysun007:
There may be a pthread_detach bug. Use pthread_join to workaround it.


guzongmin (approved):
LGTM

PR #315, opened 2021-01-15
Use goblin as elf parser and only load segments on demand 

jessehui:
Libos now Only read in headers and needed segments. Also reduce a memory
copy.

### DIFF
--- a/src/libos/src/process/do_spawn/exec_loader.rs
+++ b/src/libos/src/process/do_spawn/exec_loader.rs
@@ -1,21 +1,27 @@
+use super::super::elf_file::*;
 use super::ThreadRef;
 use crate::fs::{FileMode, INodeExt};
 use crate::prelude::*;
 use rcore_fs::vfs::INode;
 use std::ffi::CString;
 
-/// Load an ELF file itself or a script's interpreter into a vector.
+/// Load an ELF file header or a script's interpreter header into a vector.
 ///
-/// If the file is an executable binary, then just load this file.
+/// If the file is an executable binary, then just load this file's header.
 /// If the file is an script text, then parse the shebang and load
-/// the interpreter.
-pub fn load_exec_file_to_vec(
+/// the interpreter header.
+pub fn load_exec_file_hdr_to_vec(
     file_path: &str,
     current_ref: &ThreadRef,
-) -> Result<(Option<String>, Vec<u8>)> {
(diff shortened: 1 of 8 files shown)


src/libos/src/process/elf_file.rs
```diff
@@ -1,95 +1,173 @@
...
+        }
+
+        let interpreter = if intepreter_count == 0 {
+            None
+        } else {
+            str::from_utf8(&elf_buf[intepreter_offset..intepreter_offset + intepreter_count]).ok()
+        };
+        println!("interpreter = {:?}", interpreter);
+        let elf_inner =
+            goblin::elf::Elf::lazy_parse_for_occlum(header, program_headers, elf_buf, interpreter)
```
guzongmin, line 84: Is there another lazy parser? If it is the only one, we may remove the word occlum in the name.

src/libos/src/process/do_spawn/exec_loader.rs
```diff
@@ -1,21 +1,22 @@
...
-    let file_buf = load_file_to_vec(&file_path, current_ref)?;
-    let is_script = is_script_file(&file_buf);
-    if is_script {
+) -> Result<(Option<String>, Arc<dyn INode>, Vec<u8>, ElfHeader)> {
+    let (inode, file_buf, elf_hdr) = load_file_hdr_to_vec(&file_path, current_ref)?;
+    if elf_hdr.is_some() {
+        Ok((None, inode, file_buf, elf_hdr.unwrap()))
+    } else {
+        // loaded file is not Elf format, try script file
+        if !is_script_file(&file_buf) {
```
guzongmin, line 17: Is the file_buf correctly initialized in this case?

src/libos/src/process/elf_file.rs
jessehui: Removed. Thank you for your review.

src/libos/src/process/do_spawn/exec_loader.rs
jessehui: file_buf is always initialized in `read_elf64_as_vec` in deps/sefs/rcore-fs/src/vfs.rs.

jessehui:
Changes for goblin have been merged (https://github.com/m4b/goblin/pull/254) and changes for SEFS can be found [here](https://github.com/occlum/sefs/pull/11).

guzongmin (approved):
LGTM

PR #316, opened 2021-01-19
Add codecov

henrysun007:
The code coverage uses sgx_cov of Rust SGX SDK. It is ahead of the SDK version we are using, so a 885-line patch is committed together.


guzongmin (approved):
LGTM

PR #317, opened 2021-01-20
Fix demo path in CI

jessehui:


jessehui:
This is unnecessary and incorrect. We use `/root/demos` as demo path in OpenVINO and Grpc test for a reason that these two demos need specific images that have dependencies. And `/root/occlum` is mounted when the docker is created and doesn't have these dependencies in the directory.

PR #320, opened 2021-01-20
Fix gRPC and OpenVINO CI image demo location

jessehui:
The demo dir location in the CI image is wrong. This commit fixed this.


guzongmin (approved):
LGTM

PR #321, opened 2021-01-20
Fix multiplication overflow

henrysun007:


tatetian (approved):
LGTM. Thanks for the contribution.

PR #322, opened 2021-01-20
Enhance uds

henrysun007:


src/libos/src/net/socket/unix/stream/endpoint.rs
```diff
@@ -100,6 +103,42 @@ impl Inner {
...
+        if self.reader.is_peer_shutdown() {
+            events |= IoEvents::RDHUP;
+            if self.writer.is_peer_shutdown() {
+                events |= IoEvents::HUP;
+            }
+        }
+
+        let reader_events = self.reader.poll();
+        let writer_events = self.writer.poll();
+        events |= (reader_events & IoEvents::IN) | (writer_events & IoEvents::OUT);
```
tatetian, line 123: Needs experimental results to confirm that the event conversion is consistent with Linux's behavior.

src/libos/src/net/socket/unix/stream/endpoint.rs
```diff
@@ -108,3 +147,38 @@ impl Inner {
...
+        // The event of pipe should not be broadcasted directly to socket.
+        // The event transform should be consistant with poll.
+        if event.contains(IoEvents::HUP) {
+            event -= IoEvents::HUP;
+            event |= IoEvents::RDHUP;
+            if event.contains(IoEvents::ERR) {
+                event -= IoEvents::ERR;
+                event |= IoEvents::HUP;
+            }
+        }
```
tatetian, line 180: Need to confirm that the resulting event is consistent with the semantics of Linux.

src/libos/src/net/socket/unix/stream/stream.rs
```diff
@@ -273,53 +303,76 @@ impl Info {
...
+    pub fn capacity(&self) -> usize {
+        let channel = self.channel.read().unwrap();
+        channel.capacity()
+    }
+
+    pub fn resize(&self, capacity: usize) {
+        if self.capacity() == capacity {
+            return;
+        }
+
```
tatetian, line 335: Since `pop_incoming` is a blocking operation, the acquire reader lock may block the resize method for an indefinite time, which means a `listen` method (leading to `resize`) on a listening socket may be blocked by a concurrent `accept` for an indefinite time.

I don't think this issue is much of a big problem since calling `listen` more than once to adjust backlog is very rare (I don't even know about this usage before). So I think can add a TODO here to document the problem and maybe print a warning, but not fix the problem in this commit.

tatetian (commented):
I think we are really close to finish this PR. See my latest comments.

src/libos/src/net/socket/unix/stream/stream.rs
henrysun007: done

src/libos/src/fs/channel.rs
```diff
@@ -77,12 +77,32 @@ impl<I> Channel<I> {
...
     pub fn set_nonblocking(&self, nonblocking: bool) {
+        self.set_consumer_nonblocking(nonblocking);
+        self.set_producer_nonblocking(nonblocking);
+    }
+
+    pub fn set_consumer_nonblocking(&self, nonblocking: bool) {
         self.consumer.set_nonblocking(nonblocking);
+    }
+
+    pub fn set_producer_nonblocking(&self, nonblocking: bool) {
```
tatetian, line 105: Exposing these APIs is a bit of ugly. I think a more elegant solution is add `consumer` and `producer` methods so that the user of `Channel` can access them directly.

src/libos/src/net/socket/unix/stream/endpoint.rs
```diff
@@ -108,3 +141,51 @@ impl Inner {
 // TODO: Add SO_SNDBUF and SO_RCVBUF to set/getsockopt to dynamcally change the size.
 // This value is got from /proc/sys/net/core/rmem_max and wmem_max that are same on linux.
 pub const DEFAULT_BUF_SIZE: usize = 208 * 1024;
+
+/// An observer used to observe both reader and writer of the endpoint. It also contains a
+/// notifier that relays the notification of the endpoint.
+pub(super) struct RelayNotifier {
+    notifier: IoNotifier,
+    endpoint: SgxMutex<Option<Endpoint>>,
+}
```
tatetian, line 150: There is a memory leakage because `Endpoint` and `RelayNotifier` have references to each other.

henrysun007: The relationship is below:
stream(relaynotifier), stream(endpoint), relaynotifer->endpoint
Will endpoint reference relaynotifier?

tatetian: Great. Then we don't have a problem here.

src/libos/src/net/socket/unix/stream/file.rs
```diff
@@ -83,10 +90,17 @@ impl File for Stream {
...
-    fn poll(&self) -> Result<PollEventFlags> {
-        warn!("poll is not supported for unix_socket");
-        let events = PollEventFlags::empty();
-        Ok(events)
+    fn poll_new(&self) -> IoEvents {
+        match &*self.inner() {
+            // linux return value
+            Status::Idle(info) => IoEvents::OUT | IoEvents::HUP,
+            Status::Connected(endpoint) => endpoint.poll(),
+            Status::Listening(_) => IoEvents::empty(),
```
tatetian, line 98: It seems to me that the poll for a listening socket is not implemented. This is a very common functionality. I think we can add it later. But a big warning should be printed here to cry out loudly that this feature is missing.

henrysun007: It returns nothing when you poll a listener.

tatetian (approved):
LGTM. Thanks for the contribution.

PR #324, opened 2021-01-21
Add support for Address Space Layout Randomization

jessehui:
Code related with this feature is only compiled in release mode.

### DIFF
--- a/src/libos/src/vm/process_vm.rs
+++ b/src/libos/src/vm/process_vm.rs
@@ -9,6 +9,9 @@ use super::vm_manager::{
 use super::vm_perms::VMPerms;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
+// Used for heap and stack start address randomization.
+const RANGE_FOR_RANDOMIZATION: usize = 256 * 4096; // 1M
+
 #[derive(Debug, Clone)]
 pub struct ProcessVMBuilder<'a, 'b> {
     elfs: Vec<&'b ElfFile<'a>>,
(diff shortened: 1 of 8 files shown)


src/libos/src/vm/process_vm.rs
```diff
@@ -42,9 +42,38 @@ impl<'a, 'b> ProcessVMBuilder<'a, 'b> {
...
+    fn randomize_page(start_addr: usize, range: usize) -> usize {
+        use crate::util::random;
+
+        let mut random_buf: [u8; 8] = [0u8; 8]; // same length as usize
+        random::get_random(&mut random_buf).expect("failed to get random number");
+        let mut result: usize = u64::from_le_bytes(random_buf) as usize;
+        result = start_addr + result % range;
+        // make it page aligned
+        result = {
+            let rem = result % PAGE_SIZE;
```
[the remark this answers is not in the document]
jessehui: Alignment is no longer guaranteed in this function. Thank you for your review.

src/libos/src/vm/process_vm.rs
```diff
@@ -42,6 +45,21 @@ impl<'a, 'b> ProcessVMBuilder<'a, 'b> {
         self
     }
 
+    // Generate a random address within [0, range]
+    // Note: This function doesn't gurantee alignment
+    fn get_randomize_offset(range: usize) -> usize {
+        if cfg!(debug_assertions) {
+            return 0;
```
guzongmin, line 52: I think the better default implementation is "return rang;". Because if we use rang as the default, and the application works, then we could know it should work with any random value.

jessehui: Done. Thank you for your review.

src/libos/src/vm/vm_manager.rs
```diff
@@ -255,6 +256,7 @@ impl VMRemapOptions {
 pub struct VMManager {
     range: VMRange,
     vmas: Vec<VMArea>,
+    prefered_start_offset: usize, // Prefer to alloc VM starting this address offset
```
jessehui, line 259: Maybe changing this to start` address` not `offset` is better.

jessehui:
This is merged together in another PR. Thus close this one.

PR #328, opened 2021-01-26
Add unit testing framework

henrysun007:


tatetian (commented):
I am mostly positive about this PR. But I am hoping that this crate can be refactored to be decoupled from the LibOS and positions itself as a general no_std unit test framework.

The benefits of the refactoring are two-fold:
* The resulting LibOS will contain the minimal code about unit test (no `TestCase` definition, no call to inventory macros, etc.);
* The crate itself becomes more useful as it applies to other no_std projects.

# The public APIs

To be a standalone unit test crate, it should provide a user-friendly APIs that mimics the standard unit test APIs as shown below:

```rust
//! Unit test facility for no_std.
//!
//! # Example
//! 
//! ```
//! #![no_std]
//! 
//! // Instead of using `#[cfg(test)]` as we used to do in the std environment, 
//! // the unit_test crate chooses a feature named "unit_test" to conditionally compile
//! // unit test code.
//! #[cfg(feature = "unit_test")]
//! mod tests 
//! {
//! 	// Instead of using `#[test]` used to do in the std environment, the unit_test
//!	// crate choose to use an attribute named "unit_test" to specify test cases.
//!	#[unit_test]
//!	fn it_works() {
//!	    assert!(true);
//!	}
//! }
//!
//! // The user should define an entry point to run all unit tests.
//! #[cfg(feature = "unit_test")]
//! pub fn run_tests() {
//!     for test_case in unit_test::test_cases() {
//!         test_case.func();
//!     }
//! }
//! ```
```

# Changes

To implement the APIs above, the following changes to the PR are required:
* Add a crate called `unit_test` and another called `unit_test_proc` (the latter one provides the procedure macro);
* Define the `TestCase` struct in the `unit_test` crate;
* Export all `TestCase` instances through a `test_cases` method;
* Collect all `TestCase` instances by invoking `inventory::collect!` internally.

src/libos/Makefile
```diff
@@ -127,22 +132,28 @@ ifneq ($(OCCLUM_COV),)
...
+# build. It also simplifies the implementation as the release and debug build
 # have different output paths.
 ifeq ($(OCCLUM_RELEASE_BUILD), 1)
-$(LIBOS_CORE_RS_A): $(RUST_SRCS)
 	@RUSTC_BOOTSTRAP=1 RUSTC_WRAPPER=$(RUSTC_WRAPPER) cargo build --release --target-dir=$(RUST_TARGET_DIR) -Z unstable-options --out-dir=$(RUST_OUT_DIR) --features "$(LIBOS_FEATURES)"
 	@echo "CARGO (release) => $@"
 else ifneq ($(OCCLUM_COV),)
-$(LIBOS_CORE_RS_A): $(RUST_SRCS)
 	@CARGO_INCREMENTAL=0 RUSTC_BOOTSTRAP=1 RUSTFLAGS=$(COV_FLAGS) cargo build --target-dir=$(RUST_TARGET_DIR) -Z unstable-options --out-dir=$(RUST_OUT_DIR) --features "$(LIBOS_FEATURES)"
-	@echo "CARGO (debug + cov) => $@"
```
tatetian, line 140: I suppose we should keep ` + cov`.

henrysun007: I output all the features including cov and ut in the `Makefile`. Should we output `cov` here alone?

PR #336, opened 2021-02-04
Reduce redundant mprotect ocall

jessehui:
This branch is based on https://github.com/occlum/occlum/pull/324.

Previously mprotect ocall count in a make test case:
hello_world: 6
mmap: 318

With this commit (ea0aa3c), mprotect ocall count in a make test case:
hello_world: 4
mmap: 45

### DIFF
--- a/src/libos/src/vm/process_vm.rs
+++ b/src/libos/src/vm/process_vm.rs
@@ -9,6 +9,9 @@ use super::vm_manager::{
 use super::vm_perms::VMPerms;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
+// Used for heap and stack start address randomization.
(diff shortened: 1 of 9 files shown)


src/libos/src/vm/process_vm.rs
```diff
@@ -73,6 +91,7 @@ impl<'a, 'b> ProcessVMBuilder<'a, 'b> {
                     })
             })
             .collect();
+
         let other_layouts = vec![
             VMLayout::new(heap_size, PAGE_SIZE)?,
```
guzongmin, line 96: We may use a smaller alignment value in the feature implementation.  Please add a "Todo" here.

jessehui: Done. Thank you for your review.

guzongmin (approved):
LGTM

PR #340, opened 2021-02-08
Add support for mremap with file-backed memory

jessehui:
The test case covered every branch.

### DIFF
--- a/src/libos/src/vm/vm_manager.rs
+++ b/src/libos/src/vm/vm_manager.rs
@@ -7,8 +7,19 @@ use super::vm_perms::VMPerms;
 pub enum VMInitializer {
     DoNothing(),
     FillZeros(),
-    CopyFrom { range: VMRange },
-    LoadFromFile { file: FileRef, offset: usize },
+    CopyFrom {
+        range: VMRange,
+    },
+    LoadFromFile {
+        file: FileRef,
+        offset: usize,
+    },
+    // For file-backed mremap which may move from old range to new range and read extra bytes from file
+    CopyOldAndReadNew {
+        old_range: VMRange,
+        file: FileRef,
+        offset: usize, // read file from this offset
+    },
 }
 
 impl Default for VMInitializer {
@@ -42,6 +53,24 @@ impl VMInitializer {
                     .read_at(*offset, buf)
                     .cause_err(|_| errno!(EIO, "failed to init memory from file"))?;
                 for b in &mut buf[len..] {
+                    *b = 0;
+                }
+            }
+            VMInitializer::CopyOldAndReadNew {
+                old_range,
+                file,
+                offset,
+            } => {
+                // TODO: Handle old_range with non-readable subrange
+                let src_slice = unsafe { old_range.as_slice() };
+                let copy_len = src_slice.len();
+                debug_assert!(copy_len <= buf.len());
+                let read_len = buf.len() - copy_len;
+                buf[..copy_len].copy_from_slice(&src_slice[..copy_len]);
+                let len = file
+                    .read_at(*offset, &mut buf[copy_len..])
+                    .cause_err(|_| errno!(EIO, "failed to init memory from file"))?;
+                for b in &mut buf[(copy_len + len)..] {
                     *b = 0;
                 }
             }
@@ -325,7 +354,6 @@ impl VMManager {
 
         // After initializing, we can safely insert the new VMA
         self.insert_new_vma(insert_idx, new_vma);
-
         Ok(new_addr)
     }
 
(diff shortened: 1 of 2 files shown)


src/libos/src/vm/vm_manager.rs
```diff
@@ -424,14 +452,26 @@ impl VMManager {
...
 
         // Determine whether need to do mmap. And when possible, determine the returned address
-        // TODO: should fill zeros even when extending a file-backed mapping?
         let (need_mmap, mut ret_addr) = match (flags, size_type) {
             (MRemapFlags::None, SizeType::Growing) => {
+                let vm_initializer_for_new_range = if let Some((backed_file, offset)) =
+                    writeback_file
+                {
+                    // Update writeback file offset
+                    writeback_file = Some((backed_file.clone(), offset + containing_vma.size()));
```
guzongmin, line 461: May add more comments to make the logic clear.

guzongmin (commented):
LGTM

test/mmap/main.c
```diff
@@ -1091,6 +1091,135 @@ int test_mprotect_with_non_page_aligned_size() {
...
+    int byte_val_1 = 0xab;
+    int byte_val_2 = 0xcd;
+    int byte_val_3 = 0xef;
+
+    remove(file_path);
+    int fd = open(file_path, O_RDWR | O_CREAT | O_NOFOLLOW | O_CLOEXEC, 0600);
+    if (fd < 0) {
+        THROW_ERROR("open file error");
+    }
+    fallocate(fd, 0, 0, len * 4);
```
tatetian, line 1178: Why fallocate is needed? I think the following `fill_file_with_repeated_bytes` is enough.

Actually, there is a pending PR that adds the fallocate system call. So this PR doesn't even have a working implementation of fallocate right now. If fallocate is not essential, let's remove it so that the test case has fewer dependencies.

test/mmap/main.c
```diff
@@ -1091,6 +1091,135 @@ int test_mprotect_with_non_page_aligned_size() {
...
+    int prot = PROT_READ | PROT_WRITE;
+    size_t len = PAGE_SIZE;
+    char *file_path = "/tmp/test";
+    int byte_val_orig = 0xff;
+    int byte_val_1 = 0xab;
+    int byte_val_2 = 0xcd;
+    int byte_val_3 = 0xef;
+
+    remove(file_path);
+    int fd = open(file_path, O_RDWR | O_CREAT | O_NOFOLLOW | O_CLOEXEC, 0600);
```
tatetian, line 1174: The comb of `remove` + `open` can be simplified to `open(..., O_TRUNC, ...)`.

test/mmap/main.c
```diff
@@ -1091,6 +1091,135 @@ int test_mprotect_with_non_page_aligned_size() {
...
+    rc = munmap(expand_buf, 4 * len);
+    if (rc < 0) {
+        THROW_ERROR("munmap failed");
+    }
+    close(fd);
+
+    return check_file_first_four_page(file_path, byte_val_1, byte_val_2, 0, byte_val_3);
+}
+
+int test_file_backed_mremap_mem_may_move() {
```
tatetian, line 1164: The two test cases looks almost identical. Is it possible to reduce much of the code duplication with a reasonable effort of refactoring?

src/libos/src/vm/vm_manager.rs
```diff
@@ -45,6 +56,23 @@ impl VMInitializer {
                     *b = 0;
                 }
             }
+            VMInitializer::CopyOldAndReadNew {
+                old_range,
+                file,
+                offset,
+            } => {
+                let src_slice = unsafe { old_range.as_slice() };
```
tatetian, line 64: There is a possibility that the old_range has a non-readable memory permission, causing the subsequent `copy_from_slice` to panic.

src/libos/src/vm/vm_manager.rs
```diff
@@ -440,31 +480,68 @@ impl VMManager {
...
                         .build()?;
                     (Some(mmap_ops), Some(old_addr))
                 } else {
+                    // Need to move old range to a new range and init the new range
+                    let vm_initializer_for_new_range =
+                        if let Some((backed_file, offset)) = &writeback_file {
+                            VMInitializer::CopyOldAndReadNew {
+                                old_range: old_range,
+                                file: backed_file.clone(),
+                                offset: offset + containing_vma.size(),
```
tatetian, line 511: Unlike other branches, the calculation of this offset in this branch is wrong. (Hint. There is a reason that `containing_vma` is NOT named `remapped_vma`. That is, `containing_vma` contains the range `[old_addr, old_addr + old_size)`.)

src/libos/src/vm/vm_manager.rs
```diff
@@ -440,31 +480,68 @@ impl VMManager {
...
                     (Some(mmap_ops), ret_addr)
                 }
             }
             (MRemapFlags::FixedAddr(new_addr), _) => {
+                let vm_initializer_for_new_range =
+                    if let Some((backed_file, offset)) = &writeback_file {
+                        VMInitializer::CopyOldAndReadNew {
+                            old_range: old_range,
+                            file: backed_file.clone(),
+                            offset: offset + containing_vma.size(),
```
tatetian, line 534: The calculation of offset has the same problem as pointed out in the comments above.

test/mmap/main.c
jessehui: The fallocate length and `fill_file_with_repeated_bytes` are different. I think Occlum could run without fallocate but Linux will report SIGBUS because the page is not allocated yet.

src/libos/src/vm/vm_manager.rs
jessehui: Rewrite the logic with more match cases. Thank you for your review.

test/mmap/main.c
jessehui: Done.

src/libos/src/vm/vm_manager.rs
jessehui: After some digging, looks like we can't support file-backed mremap with sub range if don't support paging. I add a FIXME in the comment. Please take a look. Thank you for your revew.

tatetian (approved):
LGTM.

PR #344, opened 2021-02-20
Add support for encrypted fs image

liqinggd:
Current Occlum is using integrity-only mode or auto_key mode of SEFS, users cannot protect the image with their own key. 
This commit supports mounting an encrypted image, which is encrypted by the user key, on Occlum.

TODO:
- [x] Need to decide the API/process to launch the `init` program to mount the encrypted image on rootfs.
- Put the launch procedure of `init` program into `occlum_pal_init`.
- [x] The MAC of the same encrypted image is always changed after every`occlum build`, how to keep it constant when building the same encrypted image?
- This is because the key is randomly derived in SGX SDK, so the MAC of encrypted image changes, to keep the security as strong as SGX SDK, we do not make the MAC constant.
- [x] Refine the configuration of `mount` in Occlum.json.
- Done


liqinggd:
Remember to merge the PR in sefs https://github.com/occlum/sefs/pull/13 first.

guzongmin (dismissed):
The "encrypted fs image" is a well-designed and well-implemented feature. The only issue is there is no doc to tell the user what it is and how to use it correctly. 
It is not only the problem of this feature but also the problem of Occlum project. We should add those documents soon.

guzongmin (approved):
LGTM

joaquinboss:
> The "encrypted fs image" is a well-designed and well-implemented feature. The only issue is there is no doc to tell the user what it is and how to use it correctly. It is not only the problem of this feature but also the problem of Occlum project. We should add those documents soon.

@guzongmin 
Hello, expert. Can you look my question? I have studied Occlum recently. I have viewed official document about Occlum. But I am still confused about  encryption of the FS image. My computer is `ubuntu20.04` and `SGX_1&SGX_LC`.
The command process I understand is `mkdir occlum_instance && cd occlum_instance` `occlum init` `occlum gen-image-key image_key` `occlum build --sgx-mode HW --image-key image_key`.
Is there a problem with this process? Or will it complete encryption of FS image?
Can you look my question? Thanks a lot!

liqinggd:
Hi @joaquinboss , please refer to this https://github.com/occlum/occlum/blob/master/docs/encrypted_image.md as the short  introduction and https://github.com/occlum/occlum/tree/master/demos/remote_attestation/init_ra_flow as the real world example.

Issue #327, opened 2021-01-25
[BUG]mprotect implementation does not align with Linux


guzongmin:
# Describe the bug
mprotect syscall require a page aligned start address, but the size could be any value.
https://man7.org/linux/man-pages/man2/mprotect.2.html

In Occlum implementation, it requires the size to be page aligned too.

# Logs
[2021-01-25T03:10:19.186Z][DEBUG][T143][#1455][Mprotect] mprotect: addr: 0x7f53b7168000, size: 0x400180, perms: READ | WRITE 
[2021-01-25T03:10:19.187Z][ERROR][T143][#1455][Mprotect] Error = EINVAL (#22, Invalid argument): invalid start or end [line = 12, file = src/vm/vm_range.rs]

# Environment

- HW: [e.g. SGX1, SGX2] 
- OS: [e.g. Ubuntu18.04, CentOS8.1]
- Occlum version: [e.g. 0.19.1]

PR #345, opened 2021-02-23
Fix false alarm on non-page-aligned size for mprotect

jessehui:
Fix: https://github.com/occlum/occlum/issues/327


guzongmin (approved):
LGTM

PR #346, opened 2021-02-23
Add errno info when return error in make test

jessehui:
When encountering an error in make test and THROW_ERROR is called, output will be like this:

`ERROR:spawn process error in func test_sched_xetaffinity_with_child_pid at line 100 of file main.c with errno 12: Out of memory`


guzongmin (approved):
LGTM

PR #348, opened 2021-03-01
Fix affinity array size incorrect issue

guzongmin:
The affinity array should be compared with the CpuSet::len


tatetian (approved):
LGTM

tatetian:
@guzongmin Could you investigate why the OpenVINO test failed?

guzongmin:
The rootcause of OpenVINO failure is the lack of exception support in simulation mode.

PR #350, opened 2021-03-02
Occlum release 0.21.0

guzongmin:
1. Enable exception test support in simulation mode to help CI cover more test cases
2. Disable sccache. Because the current Rust nightly build used by Rust SDK and Occlum does not compatible. 
3. Add some workaround code in server and server_epoll test cases, because the occlum scheduling limitation.
4. Update the SGX SDK and Rust SDK to 2.13
5. Use Intel official PSW build.
6. Update the CentOS base image to 8.2.2004, becaue Intel drop the CentOS 8.1 support.
7. Handle uRTS crash case in exec mode for better debugging experience.
8. Bump the version to 0.21.0

Apply this patch when we do release 0.21.0


guzongmin:
The failed case would be recovered when the test images rebuilt with 0.21.0 base image.

tools/update_version.sh
```diff
@@ -0,0 +1,50 @@
...
+cd ../../
+
+cd src/exec/
+sed -i '/^version =/c\version = "'$OCCLUM_MAJOR_VERSION'.'$OCCLUM_MINOR_VERSION'.'$OCCLUM_PATCH_VERSION'"' Cargo.toml
+sed -i '/^name = "occlum_exec"/{n;d}' Cargo.lock
+sed -i '/^name = "occlum_exec"/a\version = "'$OCCLUM_MAJOR_VERSION'.'$OCCLUM_MINOR_VERSION'.'$OCCLUM_PATCH_VERSION'"' Cargo.lock
+cd ../../
+
+cd src/pal/include/
+cat>occlum_version.h<<EOF
```
tatetian, line 34: I think there should be a comment to mention that this header is generated automatically by this tool and should not be modified manually.

tatetian:
Needs to fix the CI tests

tools/update_version.sh
guzongmin: Updated

tatetian (approved):
Looks good.

PR #352, opened 2021-03-03
Remove HW CI test result, because the SGX machine has networking issue

guzongmin:
Temporary remove HW CI result


tatetian (approved):
LGTM

PR #355, opened 2021-03-05
Fix docker image for SGX1 HW

guzongmin:
1. Run AESM at background
2. CentOS docker image missed aesm-launch-plugin


tatetian (approved):
LGTM. But need to fix the CI failure

PR #357, opened 2021-03-08
Fix installer for new release

jessehui:
This commit fixed three errors:
(1) Fix Github action virtual environment out of free disk space
(2) Add "init" binary which is recently added for the encrypted image
(3) Bypass "Missing build-id" error when RPM packaging

This change has been verified in my local repository.

### DIFF
--- a/tools/installer/rpm/occlum.spec
+++ b/tools/installer/rpm/occlum.spec
@@ -1,8 +1,9 @@
 %define centos_base_release 1
-
 %define _unpackaged_files_terminate_build 0
 %define sgxsdk_install_dir /opt/intel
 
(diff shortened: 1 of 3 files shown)


tools/installer/rpm/occlum.spec
```diff
@@ -1,8 +1,9 @@
 %define centos_base_release 1
-
 %define _unpackaged_files_terminate_build 0
 %define sgxsdk_install_dir /opt/intel
 
+# Ignore missing build id error (generated when packaging "init" binary)
```
guzongmin, line 5: Please log an issue because Occlum user may have the same problem.

guzongmin (approved):
LGTM

PR #361, opened 2021-03-12
Remove workarounds in Demo CI

guzongmin:
Remove workarounds in Demo CI, because simulation mode has been updated.
Remove the duplicated make test.


tatetian (approved):
LGTM

PR #370, opened 2021-03-22
Export error message when parsing Elf header

jessehui:
Previously, error messages when `parse_elf_hdr ` are not exported.

Running an elf file without -fPIE compiling option will look like this:
截屏2021-03-22 下午4 45 32


guzongmin (approved):
LGTM

PR #384, opened 2021-04-07
Fix panic for non-specified exec file path

jessehui:
This commit can prevent panic when user input:
	occlum run /bin


guzongmin (approved):
LGTM

PR #404, opened 2021-04-16
Add statfs and fstatfs syscall

### LINKED ISSUE #398: Support statfs syscall
Python 3.10 will need this syscall to check if "/dev/shm" can work. 

About this syscall: https://man7.org/linux/man-pages/man2/fstatfs.2.html

liqinggd:
Fixes #398 

Remember to merge this first:
https://github.com/occlum/sefs/pull/19

### DIFF
--- a/src/libos/src/fs/file.rs
+++ b/src/libos/src/fs/file.rs
@@ -95,6 +95,10 @@ pub trait File: Debug + Sync + Send + Any {
         return_op_unsupported_error!("fallocate")
     }
 
+    fn fs(&self) -> Result<Arc<dyn FileSystem>> {
+        return_op_unsupported_error!("fs")
+    }
+
     // TODO: remove this function after all users of this code are removed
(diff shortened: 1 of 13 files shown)


src/libos/src/fs/file.rs
```diff
@@ -95,6 +95,10 @@ pub trait File: Debug + Sync + Send + Any {
         return_op_unsupported_error!("fallocate")
     }
 
+    fn fs(&self) -> Result<Arc<dyn FileSystem>> {
```
tatetian, line 98: Is it possible to return `Result<&Arc<dyn FileSystem>>`? This way, the caller can decide whether a clone on `Arc` is needed, saving unnecessary clones.

liqinggd: If return the reference, the inode_file struct have to store the Arc<dyn Filesystem> as a member. Do you think it makes sense?

tatetian (approved):
LGTM

PR #407, opened 2021-04-19
Fix ubuntu installer demo

guzongmin:


.github/workflows/package_repo_setup_and_test.yml
```diff
@@ -106,7 +106,8 @@ jobs:
 
     - name: Configure sgx and occlum deb repo
       run: |
-          docker exec ubuntu bash -c "apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends ca-certificates gnupg2 jq make gdb wget libfuse-dev libtool;
+          docker exec ubuntu bash -c "ln -snf /usr/share/zoneinfo/Asia/Shanghai /etc/localtime && echo Asia/Shanghai > /etc/timezone"
```
tatetian, line 109: Hardwiring the time zone of Shanghai does feel right to me. Could you explain this? Ideally, by adding some comments.

guzongmin: It only used to test the packages are correctly generated, so any timezone should work.

tatetian (approved):
LGTM

PR #414, opened 2021-04-27
Add pipe uid and support more detailed log for read/write

jessehui:
Current read/write log can't provide enough info for complicated applications. This commits improve this situation
1. Read/write/close syscalls will know the path or channel for regular file or pipe.
2. Read/write output a part of the buffer contents
3. Detailed log only affect debug mode

Regular file read and close: 
截屏2021-04-27 下午2 33 07

Pipe file read and close:
截屏2021-04-27 下午2 33 57

Other file types could be added in the future.


src/libos/src/fs/pipe.rs
```diff
@@ -201,4 +209,14 @@ impl PipeType for FileRef {
             .downcast_ref::<PipeWriter>()
             .ok_or_else(|| errno!(EBADF, "not a pipe writer"))
     }
+
+    fn channel_uid(&self) -> Result<usize> {
+        let pipe_end = self.as_pipe_reader();
```
guzongmin, line 214: Is it possible to get the channel, then get the channel_uid from the channel?

src/libos/src/fs/channel.rs
```diff
@@ -43,7 +45,19 @@ impl<I> Channel<I> {
...
+            consumer,
+            uid,
+        })
+    }
+
+    // Use a random number for uid
+    fn gen_uid(ring_buf: &RingBuffer<I>) -> usize {
+        use crate::util::random;
+        let mut random_buf: [u8; 8] = [0u8; 8]; // same length as usize
+        random::get_random(&mut random_buf).expect("failed to get random number");
```
guzongmin, line 59: Random does not guarantee two channels would get different id. 
It maybe a good enough implementation for debug purpose. Please add a TODO here, to replace it with a more consistent implementation.

tatetian: In case you are interested, there is an [object_id](https://github.com/occlum/ngo/blob/master/src/libos/crates/object-id/src/lib.rs) crate in NGO for exact this purpose.

jessehui: @tatetian Thank you for this suggestion. However, besides the uniqueness of this uid, it would be better if the developer can find all the references with this ID, which can be very helpful when we need to find the pattern of this access. So the uid should contain many digits. Maybe we can make `object_id` start from a big random value and then atomically increase by 1 or a small random value?

PR #421, opened 2021-04-29
Add RFC template

guzongmin:


.github/ISSUE_TEMPLATE/RFC_report.md
```diff
@@ -0,0 +1,87 @@
+---
+name: RFC template
+about: request for comments
+title: "[RFC]"
+labels: 'RFC'
+assignees: ''
+
+---
```
tatetian, line 8: The template looks fine to me.

A quick search on Google shows that this template origins from [freechipsproject/rfcs](https://github.com/freechipsproject/rfcs/blob/master/000-template.md). As a courtesy to the original authors and also an obligation to the copyright license (which is Apache License), we need to add the License file somewhere.

.github/ISSUE_TEMPLATE/RFC_report.md
```diff
@@ -0,0 +1,87 @@
...
+
+<!--- Provide a general summary of the RFC in the Title above -->
+- Feature Name: (fill me in with a unique ident, `my_awesome_feature`)
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
```
tatetian, line 12: I am really not sure what is the use of Start Date. Should there be an _End Date_?

I think maybe this info is not suitable to our needs. If we are done with discussing a RFC (whether giving it a PASS or REJECT), we can just close the issue and move on implementing and submitting PRs.

.github/ISSUE_TEMPLATE/RFC_report.md
guzongmin: This template is well-used. For example https://github.com/rust-lang/lang-team/blob/master/projects/declarative-macro-repetition-counts/0000-draft-rfc.md
I do not see such license used on those copies. So I think do not need is either.

.github/ISSUE_TEMPLATE/RFC_report.md
guzongmin: My understanding is, when we want to collect those RFC to be a document, the reader is able to know the history.

.github/ISSUE_TEMPLATE/RFC_report.md
tatetian: After doing some more googling, I found the RFC template is indeed used in several other projects. And we cannot know the original author of the text. So I guess it is ok to assume the text is in public domain and free to use.

.github/ISSUE_TEMPLATE/RFC_report.md
tatetian: Ok, make sense.

tatetian (approved):
Looks good.

PR #424, opened 2021-04-30
Add support for POSIX_SPAWN_SETSIGMASK and POSIX_SPAWN_SETSIGDEF

jessehui:


src/libos/src/process/spawn_attribute.rs
```diff
@@ -0,0 +1,97 @@
...
+// name in musl.
+#[repr(C)]
+#[derive(Debug)]
+pub struct posix_spawnattr_t {
+    flags: SpawnAttributeFlags, // __flags
+    pgrp: i32,                  // __pgrp
+    sd: SpawnAttrSigSet,        // __def
+    ss: SpawnAttrSigSet,        // __mask
+    // Below elements are different in musl but we are not using it for now. So just ignore.
+    sp: SchedParam,
```
guzongmin, line 25: If we do not use entries, should we add them now? Or can we find a better way to add both glibc and musl implementation?

jessehui: Remove the elements we don't use for now. Also, adjust the definition order to be clearer.

src/libos/src/process/do_clone.rs
```diff
@@ -59,12 +60,14 @@ pub fn do_clone(
...
             .name(name)
-            .rlimits(rlimits);
+            .rlimits(rlimits)
+            .sig_mask(sig_mask);
         if let Some(ctid) = ctid {
             builder = builder.clear_ctid(ctid);
         }
         builder.build()?
     };
+    warn!("new thread sigmask = {:?}", new_thread_ref.sig_mask());
```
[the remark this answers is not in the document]
jessehui: Done. Thanks for your review.

guzongmin (approved):
LGTM

PR #426, opened 2021-04-30
Add sendmmsg syscall

guzongmin:
sendmmsg is a Linux syscall used by glibc

### DIFF
--- a/src/libos/src/net/syscalls.rs
+++ b/src/libos/src/net/syscalls.rs
@@ -48,7 +48,7 @@ pub fn do_bind(fd: c_int, addr: *const libc::sockaddr, addr_len: libc::socklen_t
         trace!("bind to addr: {:?}", unix_addr);
         unix_socket.bind(&unix_addr)?;
     } else {
-        return_errno!(EBADF, "not a socket");
+        return_errno!(ENOTSOCK, "not a socket");
     }
 
     Ok(0)
@@ -61,7 +61,7 @@ pub fn do_listen(fd: c_int, backlog: c_int) -> Result<isize> {
     } else if let Ok(unix_socket) = file_ref.as_unix_socket() {
         unix_socket.listen(backlog)?;
     } else {
-        return_errno!(EBADF, "not a socket");
+        return_errno!(ENOTSOCK, "not a socket");
     }
 
     Ok(0)
@@ -99,7 +99,7 @@ pub fn do_connect(
 
         unix_socket.connect(&addr)?;
     } else {
-        return_errno!(EBADF, "not a socket");
+        return_errno!(ENOTSOCK, "not a socket");
     }
 
     Ok(0)
@@ -170,7 +170,7 @@ pub fn do_accept4(
         }
         Ok(new_fd as isize)
     } else {
-        return_errno!(EBADF, "not a socket");
+        return_errno!(ENOTSOCK, "not a socket");
     }
 }
 
@@ -215,7 +215,7 @@ pub fn do_setsockopt(
         warn!("setsockopt for unix socket is unimplemented");
         Ok(0)
     } else {
-        return_errno!(EBADF, "not a socket")
+        return_errno!(ENOTSOCK, "not a socket")
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/libos/src/net/syscalls.rs
```diff
@@ -531,6 +531,56 @@ pub fn do_recvmsg(fd: c_int, msg_mut_ptr: *mut msghdr_mut, flags_c: c_int) -> Re
...
+                        msg
+                    } else {
+                        return None;
+                    }
+                };
+
+                socket
+                    .sendmsg(&msg, flags)
+                    .map(|bytes_sent| {
+                        mmsg.msg_len = bytes_sent as u32;
```
tatetian, line 570: This should be adding, isn't it?
```rust
mmsg.msg_len += bytes_sent
```

src/libos/src/net/syscalls.rs
```diff
@@ -531,6 +531,56 @@ pub fn do_recvmsg(fd: c_int, msg_mut_ptr: *mut msghdr_mut, flags_c: c_int) -> Re
...
+                        mmsg.msg_len
+                    })
+                    .ok()
+            })
+            .count();
+        Ok(send_count as isize)
+    } else if let Ok(socket) = file_ref.as_unix_socket() {
+        return_errno!(EBADF, "does not support unix socket")
+    } else {
+        return_errno!(EBADF, "not a socket")
```
tatetian, line 580: The man page of `sendmsg` states:
>        EBADF  sockfd is not a valid open file descriptor.
>        ENOTSOCK The file descriptor sockfd does not refer to a socket.

So the error code here should be `ENOTSOCK`.

src/libos/src/net/syscalls.rs
```diff
@@ -531,6 +531,56 @@ pub fn do_recvmsg(fd: c_int, msg_mut_ptr: *mut msghdr_mut, flags_c: c_int) -> Re
...
+                    .map(|bytes_sent| {
+                        mmsg.msg_len = bytes_sent as u32;
+                        mmsg.msg_len
+                    })
+                    .ok()
+            })
+            .count();
+        Ok(send_count as isize)
+    } else if let Ok(socket) = file_ref.as_unix_socket() {
+        return_errno!(EBADF, "does not support unix socket")
```
tatetian, line 578: The man page of `sendmsg` states:
>        EBADF  sockfd is not a valid open file descriptor
>       EOPNOTSUPP Some bit in the flags argument is inappropriate for the socket type.

We should not use `EBADF` error code when the fd is valid, but the file type does not support this operation. The `EOPNOTSUPP` error code seems to fit our purpose better.

src/libos/src/net/syscalls.rs
guzongmin: No. It only indicates the length of single message.

src/libos/src/net/syscalls.rs
```diff
@@ -531,6 +531,56 @@ pub fn do_recvmsg(fd: c_int, msg_mut_ptr: *mut msghdr_mut, flags_c: c_int) -> Re
...
+    from_user::check_ptr(msgvec_ptr)?;
+
+    let mut msgvec = unsafe { std::slice::from_raw_parts_mut(msgvec_ptr, vlen as usize) };
+    let flags = SendFlags::from_bits_truncate(flags_c);
+    let file_ref = current!().file(fd as FileDesc)?;
+
+    if let Ok(socket) = file_ref.as_host_socket() {
+        let send_count = msgvec
+            .iter_mut()
+            .filter_map(|mut mmsg| {
```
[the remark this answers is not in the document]
guzongmin: Done

src/libos/src/net/syscalls.rs
guzongmin: Done

src/libos/src/net/syscalls.rs
guzongmin: Done

tatetian (approved):
LGTM

Issue #430, opened 2021-05-14
[BUG] Occlum-build may miscompile the Makefile under dir image/


cstwhh:
# Describe the bug
I need to package a makefile project (specifically, xgboost dependent library) to occlum instance. When I run occlum build, it will execute a command: `make -f /opt/occlum/build/bin/occlum_build.mk --always-make`, resulting in xgboost project was recompiled. This is not what I want.

# To reproduce
1. `occlum init`
1. create user Makefile:  image/Makefile:
```
test: test.cc
	@echo "Make myproject"
	gcc test.cc -o test
```
2. `occlum build`
3.  "Make myproject" will be executed

# Expected behavior
`occlum build` only does packaging and should not trigger the user's makefile to be executed

cstwhh:
'make -r' fix my problem, thanks

jessehui:
Previously, I thought the reason is the Makefile in image dir which was build again. However, it turns out that the Makefile is not the root cause, but the target executable `test` and the source file with the same name `test.c`.
Related log for `make -f /opt/occlum/build/bin/occlum_build.mk --debug=a`:
```
Considering target file '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test'.
                     Looking for an implicit rule for '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test'.
                     Trying pattern rule with stem 'test'.
                     Trying implicit prerequisite '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test.o'.
                     Trying pattern rule with stem 'test'.
                     Trying implicit prerequisite '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test.c'.
                     Found an implicit rule for '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test'.
                      Considering target file '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test.c'.
                       Looking for an implicit rule for '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test.c'.
                       Trying pattern rule with stem 'test'.
                       Trying implicit prerequisite '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test.y'.
                       Trying pattern rule with stem 'test'.
                       Trying implicit prerequisite '/root/Dev/SHARE/occlum_build_helper/occlum_workspace/image/lib/test.l'.
[...]

PR #431, opened 2021-05-14
Fix occlum build "-f" option mistakenly rebuild other targets

jessehui:
Remove "--always-remake" option and add "--no-builtin-rules" to avoid
mistakenly rebuild unwanted targets.

Fixed issue #430


guzongmin (approved):
LGTM

PR #443, opened 2021-05-25
Check AESM status with aesm socket file

guzongmin:
Some user using host AESM, so Occlum should only check the socket file instead of the AESM process before running.


jessehui (approved):
LGTM

tatetian (approved):
LGTM

PR #446, opened 2021-05-26
Fix locatime & quote verify shared lib

qiyuangong:
* Add tzdata & libsgx-dcap-quote-verify-dev into install package


guzongmin (approved):
LGTM

Issue #429, opened 2021-05-14
[RFC] Support execve syscall


jessehui:
- Feature Name: support execve syscall
- Start Date: 2021-5-12

# Summary
[summary]: #summary
Support execve syscall in Occlum.

# Motivation
[motivation]: #motivation
`Execve` is a very important system call that is widely used by system-level applications. It usually works with fork together when creating new processes. Occlum use spawn rather than fork+exec to create a new process. But in some scenarios, execve syscall could be used alone. 

During the enablement of Java multiprocessing, we can hardcode the method of creating a new process to spawn based on:
![截屏2021-05-12 下午2 50 09](https://user-images.githubusercontent.com/13930406/117931113-5e317100-b331-11eb-8ef8-0bad23f4c148.png)

However, although it is set to use posix_spawn, it actually spawns a `spawnhelper` defined in `src/java.base/unix/native/jspawnhelper/jspawnhelper.c` and calls `execve` later to execute the real child process. Thus, `execve` becomes a "must" to enable Java multiprocessing.

What's more, shell command `exec` could also depend on this.

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

In general, it follows the beviour of execve described in the [man page](https://man7.org/linux/man-pages/man2/execve.2.html). Besides execve a elf binary, it should also support execve a script.

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

In native Linux, execve syscall should replace all the VM of a process and execute the new process. However, for single address space OS, we can't do that. A reasonable method is to create a new process in the existing way. And then just execute the new process in the syscall and never return. In more details:
1. `process 1` (main process) calls execve
2. OS handles this syscall and creates a new process, let's say `process 1'`, similar to spawn but in a replace way:
      - inherit the `process 1`'s parent process as its parent process
      - use the same tid and pid as `process 1` while replacing process reference and thread reference in the process table
      - close all the `close-on-exec` file and others are kept
      - the dispositions of handled signals are reset to the default; the dispositions of ignored signals are left unchanged
      - any alternate signal stack is not preserved
      - any open directory streams are closed
      - any outstanding asynchronous I/O operations are canceled
      - all `process 1`'s threads are destroyed during an execve().  Mutexes, condition variables, and other pthreads objects are not preserved.
3. after creating `process 1'`, using OCALL to run it in a host thread
4. exit the original `process 1` including terminating the thread and release the resources of process and thread based on its reference

# Rationale and alternatives
[...]

PR #448, opened 2021-05-26
Add support for execve

jessehui:
RFC: #429 
This is based on #424.

Known Limitations:

1. `occlum exec` command will return early if the application calls execve. Because
the "execved"-ed application will run on a new thread and the current thread will exit.
`occlum run` will not have this problem.

2. GDB doesn't work well for "execve"-ed application. Because execve will reuse the
virtual memory and there is no API for unloading previous symbols.
After do_exec is called, users can use `remove-symbol-file` command in GDB to unload previous symbols by hand.

3. Based on the man page, "execve" should return -1, and set errno to indicate the error.
However, in this implementation, "execve" doesn't return with errno. Instead, the current
thread directly exits with the errno as termination status. Thus, the process will just exit.


src/libos/src/process/do_spawn/mod.rs
```diff
@@ -97,8 +105,68 @@ fn new_process(
...
+    Ok(new_process_ref)
+}
+
+/// Create a new process for execve which will use same parent, pid, tid and reuse VM if possible
+pub fn new_process_for_exec(
+    file_path: &str,
+    argv: &[CString],
+    envp: &[CString],
+    current_ref: &ThreadRef,
+    current_pid: pid_t,
```
guzongmin, line 135: The current_pid is not needed. current_pid is part of current_ref.

jessehui:
In the second version, execve will allocate new address space to align with Linux (and maybe to support vfork in the future). And instead of exiting all threads at the entrance of the syscall, the current process will "freeze" to wait for the exec status and only exit when exec succeeds.

Because of this, the limitation 2 and 3 will no longer exist.

src/libos/src/process/process/mod.rs
```diff
@@ -123,12 +128,37 @@ impl Process {
...
     }
 
+    /// Difference with freeze defined in process:
+    /// - force_freeze only changes the thread status
+    /// - freeze element in process struct achieves the "freeze" behaviour
+    pub fn force_freeze(&self, tid: pid_t) {
+        self.forced_exit_status.force_freeze(tid);
+        let threads = self.threads();
+        for thread in threads.iter() {
+            if thread.tid() == tid {
```
guzongmin, line 145: Do not get the point. Why do you freeze only one thread?

src/libos/src/process/process/mod.rs
```diff
@@ -123,12 +128,37 @@ impl Process {
...
+            }
+            info!("force thread {:?} to freeze", thread.tid());
+            thread.inner().freeze()
+        }
+    }
+
+    /// Set all the thread to running status
+    pub fn force_unfreeze(&self) {
+        let threads = self.threads();
+        for thread in threads.iter() {
```
jessehui, line 156: use map instead of for

src/libos/src/process/process/mod.rs
jessehui: Updated to freeze all threads including calling thread.

src/libos/src/process/do_exec.rs
```diff
@@ -0,0 +1,86 @@
...
+        // 4. Force exit all child threads of current process
+        // Once this succeeds, unfreeze and directly exit all threads.
+        let term_status = TermStatus::Exited(0 as u8);
+        current.process().force_exit(term_status);
+        // There might be some threads freezed at the sysret and wait for new status.
+        // Unfreeze these threads.
+        *freeze = false;
+        drop(freeze);
+        // Don't hesitate. Interrupt all threads right now (except the calling thread).
+        broadcast_interrupts();
```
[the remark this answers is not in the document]
jessehui: `broadcast_interrupt` should be called after force_exit and the threads will be interrupted and handled by `handle_force_exit`.

src/libos/src/process/do_exec.rs
```diff
@@ -0,0 +1,86 @@
...
+        current_ref.process().pid()
+    );
+
+    // 1. Freeze all child threads of current process to wait for exec status
+    let current = current!();
+    // 1.1 This must be locked to freeze other threads
+    let mut freeze = current.process().freeze();
+    *freeze = true;
+    // 1.2 Set process status to freeze
+    current.process().force_freeze(current_ref.tid());
```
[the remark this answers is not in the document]
jessehui: Done. thanks for your review.

src/libos/src/process/do_exec.rs
```diff
@@ -0,0 +1,86 @@
...
+        // Unfreeze these threads.
+        *freeze = false;
+        drop(freeze);
+        // Don't hesitate. Interrupt all threads right now (except the calling thread).
+        broadcast_interrupts();
+
+        // 5. Wait for all threads (except calling thread) to exit
+        let mut thread_num = current.process().threads().len();
+        while (thread_num != 1) {
+            thread_num = current.process().threads().len();
```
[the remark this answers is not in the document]
jessehui: The calling thread must be waiting here because it has to guarantee that the calling thread is the last exit thread to do some extra work. However, I didn't find out an easy way for the calling thread to wait for other threads to exit and not occupying the CPU. Added FIXME here.

src/libos/src/process/process/mod.rs
jessehui: Done

src/libos/src/process/do_spawn/mod.rs
```diff
@@ -107,6 +108,59 @@ fn new_process(
...
+    let tid = ThreadId {
+        tid: current_ref.process().pid() as u32,
+    };
+    let new_process_ref = new_process_common(
+        file_path,
+        argv,
+        envp,
+        &Vec::new(),
+        None,
+        None,
```
guzongmin, line 145: Can you please check the stdio status?

src/libos/src/process/do_spawn/mod.rs
```diff
@@ -223,10 +277,21 @@ fn new_process(
         let elf_name = elf_path.rsplit('/').collect::<Vec<&str>>()[0];
         let thread_name = ThreadName::new(elf_name);
 
-        ProcessBuilder::new()
+        let mut parent;
+        let mut process_builder = ProcessBuilder::new();
+        if reuse_tid.is_some() {
+            // must be in execve syscall
+            // assert!(reuse_vm.is_some());
```
guzongmin, line 284: Enable the assertion or remove it

src/libos/src/process/do_spawn/mod.rs
jessehui: Based on `do_spawn`, the host_stdio_fd should be set to `None` because it should have been set by `occlum_ecall_new_process` at the Occlum entry.

src/libos/src/process/do_spawn/mod.rs
jessehui: This assertion is not correct anymore. Removed.

src/libos/src/vm/vm_manager.rs
```diff
@@ -326,6 +326,11 @@ impl VMManager {
         self.mmap_prefered_start_addr = addr
     }
 
+    // This is used to clean all vmas
+    pub fn sanitize(&mut self) -> Result<()> {
+        self.munmap(self.range.start(), self.range().size())
```
guzongmin, line 331: self.range.start() could be self.range().start()

src/libos/src/vm/user_space_vm.rs
```diff
@@ -109,6 +109,18 @@ impl UserSpaceVMRange {
     pub fn range(&self) -> &VMRange {
         &self.vm_range
     }
+
+    pub fn sanitize(&self) {
```
guzongmin, line 113: The sanitize function maybe not needed.

src/libos/src/vm/process_vm.rs
```diff
@@ -108,17 +108,73 @@ impl<'a, 'b> ProcessVMBuilder<'a, 'b> {
...
+        // preserving extra space for alignment
+        let new_process_size = process_layout.align() + process_layout.size();
+
+        // Reuse old VM if possible
+        if let Some(old_vm) = reuse_process_vm {
+            let old_range = old_vm.process_range().range();
+            if old_range.size() >= new_process_size {
+                let mut vm_manager = old_vm.vm_manager();
+                // Clean VM manager
+                vm_manager.sanitize()?;
```
guzongmin, line 121: If the reuse case "vfork", we should not sanitize the vm_manager.

src/libos/src/vm/vm_manager.rs
jessehui: This part is useless. Removed.

src/libos/src/vm/user_space_vm.rs
jessehui: Removed.

src/libos/src/vm/process_vm.rs
jessehui: True. Removed.

guzongmin (approved):
LGTM

guzongmin (approved):
LGTM

PR #453, opened 2021-06-01
Fix a potential deadlock when dereference procfs's symlink

liqinggd:
## Problem:
When process lookup the inode through `lookup_inode` of FsView, It has to hold the lock of FsView, meanwhile, the symlink inode of procfs may access the `cwd` or `root` of FsView when dereferencing symlinks. Apparently, there is a deadlock of FsView.

## Solution:
Change the lock of FsView from `SgxMutex` to `RwLock`, to access the `cwd` and `root` of FsView concurrently.


guzongmin (approved):
LGTM

PR #454, opened 2021-06-01
Fix unix domain socket when enabling Java UT

jessehui:
Include two commits:
- Add support for indexing unix domain socket file with inode
- Add dummy implementation of getsockopt for unix domain socket

### DIFF
--- a/src/libos/src/net/socket/unix/addr.rs
+++ b/src/libos/src/net/socket/unix/addr.rs
@@ -10,7 +10,7 @@ lazy_static! {
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Addr {
-    File(UnixPath),
+    File(Option<usize>, UnixPath), // An optional inode number and path. Use inode if there is one.
     Abstract(String),
 }
 
@@ -54,7 +54,7 @@ impl Addr {
                 return_errno!(EINVAL, "no null in the address");
             }
 
-            Ok(Self::File(UnixPath::new(path_cstr.to_str().unwrap())))
+            Ok(Self::File(None, UnixPath::new(path_cstr.to_str().unwrap())))
         }
     }
 
@@ -76,7 +76,7 @@ impl Addr {
 
     pub fn path_str(&self) -> &str {
         match self {
-            Self::File(unix_path) => &unix_path.path_str(),
+            Self::File(_, unix_path) => &unix_path.path_str(),
             Self::Abstract(path) => &path,
         }
     }
@@ -86,7 +86,7 @@ impl Addr {
         addr.sun_family = AddressFamily::LOCAL as libc::sa_family_t;
 
         let addr_len = match self {
-            Self::File(unix_path) => {
+            Self::File(_, unix_path) => {
                 let path_str = unix_path.path_str();
                 let buf_len = path_str.len();
                 /// addr is initialized to all zeros and try_from_raw guarentees
--- a/src/libos/src/net/socket/unix/stream/address_space.rs
+++ b/src/libos/src/net/socket/unix/stream/address_space.rs
@@ -7,9 +7,27 @@ lazy_static! {
     pub(super) static ref ADDRESS_SPACE: AddressSpace = AddressSpace::new();
 }
 
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+pub enum AddressSpaceKey {
+    FileKey(usize),
+    AbstrKey(String),
+}
+
+impl AddressSpaceKey {
+    pub fn from_inode(inode: usize) -> Self {
+        AddressSpaceKey::FileKey(inode)
+    }
+
+    pub fn from_path(path: String) -> Self {
+        AddressSpaceKey::AbstrKey(path)
+    }
+}
+
 pub struct AddressSpace {
-    file: SgxMutex<BTreeMap<String, Option<Arc<Listener>>>>,
-    abstr: SgxMutex<BTreeMap<String, Option<Arc<Listener>>>>,
+    // For "file", use inode number as "key" instead of path string so that listeners can still
+    // be reached even if the socket file is moved or renamed.
+    file: SgxMutex<BTreeMap<AddressSpaceKey, Option<Arc<Listener>>>>,
+    abstr: SgxMutex<BTreeMap<AddressSpaceKey, Option<Arc<Listener>>>>,
 }
 
 impl AddressSpace {
@@ -21,7 +39,7 @@ impl AddressSpace {
     }
 
     pub fn add_binder(&self, addr: &Addr) -> Result<()> {
-        let key = Self::get_key(addr);
+        let key = Self::get_key(addr).ok_or_else(|| errno!(EINVAL, "can't find socket file"))?;
         let mut space = self.get_space(addr);
         if space.contains_key(&key) {
             return_errno!(EADDRINUSE, "the addr is already bound");
@@ -32,7 +50,7 @@ impl AddressSpace {
     }
 
     pub fn add_listener(&self, addr: &Addr, capacity: usize, nonblocking: bool) -> Result<()> {
-        let key = Self::get_key(addr);
+        let key = Self::get_key(addr).ok_or_else(|| errno!(EINVAL, "the socket is not bound"))?;
         let mut space = self.get_space(addr);
 
         if let Some(option) = space.get(&key) {
@@ -48,7 +66,7 @@ impl AddressSpace {
     }
 
     pub fn resize_listener(&self, addr: &Addr, capacity: usize) -> Result<()> {
-        let key = Self::get_key(addr);
+        let key = Self::get_key(addr).ok_or_else(|| errno!(EINVAL, "the socket is not bound"))?;
         let mut space = self.get_space(addr);
 
         if let Some(option) = space.get(&key) {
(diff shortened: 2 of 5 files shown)


src/libos/src/net/socket/unix/stream/address_space.rs
```diff
@@ -88,17 +107,39 @@ impl AddressSpace {
...
+    ) -> SgxMutexGuard<'_, BTreeMap<AddressSpaceKey, Option<Arc<Listener>>>> {
         match addr {
-            Addr::File(unix_path) => self.file.lock().unwrap(),
+            Addr::File(unix_path, inode) => self.file.lock().unwrap(),
             Addr::Abstract(path) => self.abstr.lock().unwrap(),
         }
     }
 
-    fn get_key(addr: &Addr) -> String {
+    fn get_key(addr: &Addr) -> AddressSpaceKey {
```
liqinggd, line 120: Rewrite this function

```
fn get_key(addr: &Addr) -> Option<AddressSpaceKey> {
    trace!("addr = {:?}", addr);
    match addr {
        Addr::File(unix_path, inode_num) if inode_num.is_some() => {
            Some(AddressSpaceKey::from_inode(inode_num.unwrap()))
        }
        Addr::File(unix_path, _) => {
            let inode = {
                let file_path = unix_path.absolute();
                let current = current!();
                let fs = current.fs().read().unwrap();
                fs.lookup_inode(&file_path)
            };
            if let Ok(inode) = inode {
                Some(AddressSpaceKey::from_inode(inode.metadata().unwrap().inode));
            } else {
                None
            }
        }
        Addr::Abstract(path) => Some(AddressSpaceKey::from_path(addr.path_str().to_string())),
    }
}
```

src/libos/src/net/socket/unix/stream/address_space.rs
```diff
@@ -7,9 +7,28 @@ lazy_static! {
...
 
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+pub enum AddressSpaceKey {
+    FileKey(usize),
+    AbstrKey(String),
+    NoExist,
+}
+
+impl AddressSpaceKey {
+    pub fn new(file_key: Option<usize>, abstr_key: Option<String>) -> Self {
```
liqinggd, line 18: Separate the `new` into `AddressSpaceKey::from_inode` and `AddressSpaceKey::from_path` to make it more clear for the initialization of it.

src/libos/src/net/socket/unix/stream/address_space.rs
```diff
@@ -7,9 +7,28 @@ lazy_static! {
     pub(super) static ref ADDRESS_SPACE: AddressSpace = AddressSpace::new();
 }
 
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
+pub enum AddressSpaceKey {
+    FileKey(usize),
+    AbstrKey(String),
+    NoExist,
```
liqinggd, line 14: `NoExsit` seems to be useless, should we remove it and use `None` for the return value of `get_key`? If `get_key` returns `None`, should we return the Error?

src/libos/src/net/socket/unix/addr.rs
```diff
@@ -86,7 +86,7 @@ impl Addr {
         addr.sun_family = AddressFamily::LOCAL as libc::sa_family_t;
 
         let addr_len = match self {
-            Self::File(unix_path) => {
+            Self::File(unix_path, inode_num) => {
```
liqinggd, line 89: the second parameter is useless, why not mark it with `Self::File(unix_path, _)`

src/libos/src/net/socket/unix/stream/stream.rs
```diff
@@ -68,8 +69,24 @@ impl Stream {
...
     }
 
-    // TODO: create the corresponding file in the fs
-    pub fn bind(&self, addr: &Addr) -> Result<()> {
+    pub fn bind(&self, addr: &mut Addr) -> Result<()> {
+        let inode_num = if let Addr::File(path, inode) = addr {
+            // create the corresponding file in the fs and fill Addr with its inode
+            let current = current!();
+            let fs = current.fs().read().unwrap();
+            let file = fs.open_file(&addr.path_str(), CreationFlags::O_CREAT.bits(), 0x777)?;
```
liqinggd, line 77: should be 0o777, not 0x777

src/libos/src/net/socket/unix/stream/stream.rs
```diff
@@ -68,8 +69,24 @@ impl Stream {
         return_errno!(ENOTCONN, "the socket is not connected");
     }
 
-    // TODO: create the corresponding file in the fs
-    pub fn bind(&self, addr: &Addr) -> Result<()> {
+    pub fn bind(&self, addr: &mut Addr) -> Result<()> {
+        let inode_num = if let Addr::File(path, inode) = addr {
```
liqinggd, line 73: Rewrite it with clear logic
```
if let Addr::File(path, inode_num) = addr {
        // create the corresponding file in the fs and fill Addr with its inode
        let corresponding_inode_num = {
            let current = current!();
            let fs = current.fs().read().unwrap();
            let file_ref = fs.open_file(&addr.path_str(), CreationFlags::O_CREAT.bits(), 0o777)?;
            file_ref.metadata()?.inode
        };
        *inode_num = Some(corresponding_inode_num);
    }
```

src/libos/src/net/socket/unix/addr.rs
```diff
@@ -10,7 +10,7 @@ lazy_static! {
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Addr {
-    File(UnixPath),
+    File(UnixPath, Option<usize>), // path or inode
```
liqinggd, line 13: // path and an optional inode number

liqinggd (changes requested):

src/libos/src/net/socket/unix/stream/stream.rs
jessehui: Done.

src/libos/src/net/socket/unix/stream/address_space.rs
jessehui: Thank you for the review. I think this is more concise and clear. Done.

src/libos/src/net/socket/unix/addr.rs
jessehui: Done.

src/libos/src/net/socket/unix/addr.rs
jessehui: Done.

src/libos/src/net/socket/unix/stream/address_space.rs
jessehui: removed.

src/libos/src/net/socket/unix/stream/address_space.rs
jessehui: Done.

src/libos/src/net/socket/unix/stream/stream.rs
tatetian: @liqinggd You got a sharp eye 👍

jessehui (commented):
Thanks!

src/libos/src/net/socket/unix/addr.rs
tatetian: I am wondering why using a pair of `UnixPath` and `Option<usize>` instead of an `InodeFile`? Essentially, a pathname-based UDS should be backed by an inode. An inode provides more info than the combination of path and inode number.

test/unix_socket/main.c
```diff
@@ -65,6 +65,67 @@ int create_connceted_sockets_default(int *sockets) {
...
+    int listen_fd = socket(AF_UNIX, SOCK_STREAM, 0);
+    if (listen_fd == -1) {
+        THROW_ERROR("failed to create a unix socket");
+    }
+
+    struct sockaddr_un addr;
+    memset(&addr, 0, sizeof(struct sockaddr_un)); //Clear structure
+    addr.sun_family = AF_UNIX;
+    strcpy(addr.sun_path, socket_original_path);
+    socklen_t addr_len = strlen(addr.sun_path) + sizeof(addr.sun_family);
```
tatetian, line 80: I suppose that `addr_len` should be one byte longer according to the man page:

```
       Three types of address are distinguished in the sockaddr_un
       structure:

       *  pathname: a UNIX domain socket can be bound to a null-
          terminated filesystem pathname using bind(2).  When the
          address of a pathname socket is returned (by one of the system
          calls noted above), its length is

              offsetof(struct sockaddr_un, sun_path) + strlen(sun_path)
          + 1
```

Other locations of address length calculation in the test should also be fixed.

test/unix_socket/main.c
```diff
@@ -65,6 +65,67 @@ int create_connceted_sockets_default(int *sockets) {
...
+        THROW_ERROR("failed to listen");
+    }
+
+    // rename to new path
+    unlink(socket_ready_path);
+    if (rename(socket_original_path, socket_ready_path) < 0) {
+        THROW_ERROR("failed to rename");
+    }
+
+    int client_fd = socket(AF_UNIX, SOCK_STREAM, PF_UNIX);
```
tatetian, line 97: The third argument `protocol` of the `socket` API should be given a value of 0, indicating that you want the default protocol. There is no other protocols available in the unix socket family.




Here is the relevant paragraphs in man pages:

> Normally only a single protocol exists to support a
       particular socket type within a given protocol family, in which
       case protocol can be specified as 0.

> #include <sys/socket.h>
> #include <sys/un.h>
> unix_socket = socket(AF_UNIX, type, 0);

> The manifest constants used under 4.x BSD for protocol families
> are PF_UNIX, PF_INET, and so on, while AF_UNIX, AF_INET, and so
> on are used for address families.  However, already the BSD man
>  page promises: "The protocol family generally is the same as the
>  address family", and subsequent standards use AF_* everywhere.

And `PF_UNIX` is just an alias of `AF_UNIX` or `AF_LOCAL`. So both `PF_UNIX` and `AF_UNIX` are meant to be used for the first argument of `family`. 

```c
// File include/sys/socket.h in musl libc

#define PF_LOCAL        1
#define PF_UNIX         PF_LOCAL
#define PF_FILE         PF_LOCAL
//...

#define AF_LOCAL        PF_LOCAL
#define AF_UNIX         AF_LOCAL
#define AF_FILE         AF_LOCAL
//...
```

src/libos/src/net/socket/unix/addr.rs
jessehui: The reason is adding an optional parameter is an easy fix and will not affect existing code and function.

test/unix_socket/main.c
jessehui: Got it. Done. Added a comment for notation.

test/unix_socket/main.c
jessehui: Done. Thanks for your review.

src/libos/src/net/socket/unix/stream/stream.rs
liqinggd: Then why not clone the `addr.path_str()` to a String and pass the reference of the newly cloned String to `open_file`?

liqinggd: Hi, I think the code below can work

```
if let Addr::File(inode_num, unix_path) = addr {
            // create the corresponding file in the fs and fill Addr with its inode
            let corresponding_inode_num = {
                let current = current!();
                let fs = current.fs().read().unwrap();
                let file_ref =
                    fs.open_file(&unix_path.path_str(), CreationFlags::O_CREAT.bits(), 0o777)?;
                file_ref.metadata()?.inode
            };
            *inode_num = Some(corresponding_inode_num);
        }
```

jessehui: This logic is more clear. Changed a little to make it work. Thanks for your review.

jessehui: Yes, mine is the same. Please take a look. Thank you!

liqinggd:
LGTM for my comments, please wait for the feedback from @tatetian

guzongmin (approved):
LGTM

PR #475, opened 2021-06-11
Fix dockerfile error

jessehui:
1. Make "bash" the default shell when building image
2. Add libarchive dependency for centos image


guzongmin (approved):
LGTM

Issue #422, opened 2021-04-29
[RFC] Generate a secure /etc/resolv.conf file for Occlum


guzongmin:
- Feature Name: generate_secure_resolv_config
- Start Date: 2012-04-29

# Summary
[summary]: #summary

Generate a secure /etc/resolv.conf at the Occlum boot time.

# Motivation
[motivation]: #motivation

The software running inside Occlum need to get the DNS information by reading /etc/resolv.conf. So Occlum need to provide this file to user. But this file changes time by time. As the result Occlum need to create this file on the fly. Copy the host /etc/resolv.conf into occlum is a simple solution, but it is not a secure way, because someone may replace it with another file.

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

The occlum user would not aware of it, it is totally transparent.
If someone copies the /etc/resolv.conf into occlum in the existing cases, then he/she would free of this task. If the user want to use a special version resolv.conf, just copy that file into /etc/. Occlum would not replace the file, if it already exists.

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

The /etc/resolv.conf file is a configuration file. There are some parsers to read the content from this file. https://github.com/tailhook/resolv-conf is a reference parser implementation.

AT the occlum boot time, if the /ete/resolv.conf file in occlum is not presented, occlum need to generate one for user. 
1. Reads the host machine /etc/resolv.conf file.
2. Parse this file to verify it is a correct configuration file.
3. Copy/generate the /etc/resolv.conf inside occlum

# Drawbacks
[drawbacks]: #drawbacks

Is it secure enough?

# Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

Alternative 1: Copy the /etc/resolv.conf when build the occlum image.
+ It is not a good solution, because the DNS would change.

Alternative 2: Copy the /etc/resolv.conf at boot time without any check.
+ There is a security issue, if someone replaces it with a malware.

Occlum gets some issues related with this feature: #420 #415 #202

# Future possibilities
[future-possibilities]: #future-possibilities

If Occlum needs to copy any file into occlum image at runtime, it should check before copy as this design.

PR #482, opened 2021-06-22
Add resolv.conf parser

ClawSeven:
Add the resolv.conf parser. #422

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "deps/ringbuf"]
 	path = deps/ringbuf
 	url = https://github.com/agerasev/ringbuf.git
+[submodule "deps/resolv-conf"]
+	path = deps/resolv-conf
+	url = https://github.com/tailhook/resolv-conf.git
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ submodule: githooks
 	@# Try to apply the patches. If failed, check if the patches are already applied
 	cd deps/serde-json-sgx && git apply ../serde-json-sgx.patch >/dev/null 2>&1 || git apply ../serde-json-sgx.patch -R --check
 	cd deps/ringbuf && git apply ../ringbuf.patch >/dev/null 2>&1 || git apply ../ringbuf.patch -R --check
+	cd deps/resolv-conf && git apply ../resolv-conf.patch >/dev/null 2>&1 || git apply ../resolv-conf.patch -R --check
 
 	@# Enclaves used by tools are running in simulation mode by default to run faster.
 	@rm -rf build build_sim
--- a/src/libos/src/entry.rs
+++ b/src/libos/src/entry.rs
@@ -12,6 +12,7 @@ use crate::signal::SigNum;
 use crate::time::up_time::init;
 use crate::util::log::LevelFilter;
 use crate::util::mem_util::from_untrusted::*;
+use crate::util::resolv_conf_util::{parse_resolv_conf, write_resolv_conf};
 use crate::util::sgx::allow_debug as sgx_allow_debug;
 use sgx_tse::*;
 
@@ -23,6 +24,7 @@ lazy_static! {
     static ref HAS_INIT: AtomicBool = AtomicBool::new(false);
     pub static ref ENTRY_POINTS: RwLock<Vec<PathBuf>> =
         RwLock::new(config::LIBOS_CONFIG.entry_points.clone());
+    pub static ref RESOLV_CONF_STR: RwLock<Option<String>> = RwLock::new(None);
 }
 
 macro_rules! ecall_errno {
@@ -33,7 +35,11 @@ macro_rules! ecall_errno {
 }
 
 #[no_mangle]
-pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *const c_char) -> i32 {
+pub extern "C" fn occlum_ecall_init(
+    log_level: *const c_char,
+    instance_dir: *const c_char,
+    resolv_conf_ptr: *const c_char,
+) -> i32 {
     if HAS_INIT.load(Ordering::SeqCst) == true {
         return ecall_errno!(EEXIST);
     }
@@ -86,6 +92,18 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
         // Enable global backtrace
         unsafe { backtrace::enable_backtrace(&ENCLAVE_PATH, PrintFormat::Short) };
     });
+
+    match parse_resolv_conf(resolv_conf_ptr) {
+        Err(e) => {
+            error!("failed to parse /etc/resolv.conf: {}", e.backtrace());
+        }
+        Ok(resolv_conf_str) => {
+            *RESOLV_CONF_STR.write().unwrap() = Some(resolv_conf_str);
+            if let Err(e) = write_resolv_conf() {
+                error!("failed to write /etc/resolv.conf: {}", e.backtrace());
+            }
+        }
+    }
 
     0
 }
--- a/src/libos/src/fs/fs_ops/mount.rs
+++ b/src/libos/src/fs/fs_ops/mount.rs
@@ -1,4 +1,5 @@
 use std::sync::Once;
+use util::resolv_conf_util::write_resolv_conf;
 
 use super::rootfs::{mount_nonroot_fs_according_to, open_root_fs_according_to};
 use super::*;
@@ -27,5 +28,9 @@ pub fn do_mount_rootfs(
         *root_inode = new_root_inode;
         *ENTRY_POINTS.write().unwrap() = user_config.entry_points.to_owned();
     });
+    // Write resolv.conf file into mounted file system
+    write_resolv_conf()?;
+    *RESOLV_CONF_STR.write().unwrap() = None;
+
     Ok(())
 }
--- a/src/libos/src/lib.rs
+++ b/src/libos/src/lib.rs
@@ -53,6 +53,7 @@ extern crate serde;
 extern crate serde_json;
 #[macro_use]
 extern crate memoffset;
+extern crate resolv_conf;
 
 use sgx_trts::libc;
 use sgx_types::*;
--- a/src/libos/src/util/mod.rs
+++ b/src/libos/src/util/mod.rs
@@ -5,5 +5,6 @@ pub mod log;
 pub mod mem_util;
 pub mod mpx_util;
 pub mod random;
+pub mod resolv_conf_util;
 pub mod sgx;
 pub mod sync;
--- a/src/libos/src/util/resolv_conf_util.rs
+++ b/src/libos/src/util/resolv_conf_util.rs
@@ -1,0 +1,37 @@
+use super::*;
+use crate::fs::{AccessMode, CreationFlags, FsView};
+use resolv_conf::*;
+use std::ffi::CStr;
+use std::str;
+
+pub fn write_resolv_conf() -> Result<()> {
+    const RESOLV_CONF_PATH: &'static str = "/etc/resolv.conf";
+    let fs_view = FsView::new();
+    // overwrite /etc/resolv.conf if existed
+    let resolv_conf_file = fs_view.open_file(
+        RESOLV_CONF_PATH,
+        AccessMode::O_RDWR as u32 | CreationFlags::O_CREAT.bits() | CreationFlags::O_TRUNC.bits(),
+        0o666,
+    )?;
+    let resolv_conf_str = RESOLV_CONF_STR.read().unwrap();
+    match &*resolv_conf_str {
+        Some(str) => {
+            resolv_conf_file.write(str.as_bytes());
+        }
+        None => {}
+    }
+    Ok(())
+}
+
+pub fn parse_resolv_conf(resolv_conf_ptr: *const c_char) -> Result<String> {
+    // Read resolv.conf file from host
+    let resolv_conf_bytes = unsafe { CStr::from_ptr(resolv_conf_ptr).to_bytes() };
+    let resolv_conf_str = str::from_utf8(resolv_conf_bytes)
+        .map_err(|_| errno!(EINVAL, "/etc/resolv.conf contains non UTF-8 characters"))?;
+
+    // Parse and inspect resolv.conf file
+    if let Err(_) = resolv_conf::Config::parse(resolv_conf_bytes) {
+        return_errno!(EINVAL, "malformated host /etc/resolv.conf");
+    }
+    Ok(resolv_conf_str.to_string())
+}
--- a/src/pal/src/pal_load_resolv_conf.c
+++ b/src/pal/src/pal_load_resolv_conf.c
@@ -1,0 +1,24 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "pal_log.h"
+
+char *pal_load_resolv_conf(void) {
+    FILE *fp = fopen("/etc/resolv.conf", "rb");
+
+    if (fp == NULL) {
+        PAL_WARN("Warning: Failed to open /etc/resolv.conf file");
+        return NULL;
+    }
+    fseek(fp, 0, SEEK_END);
+    long fsize = ftell(fp);
+    fseek(fp, 0, SEEK_SET);
+    char *resolv_conf_buffer = malloc(fsize + 1);
+    if (resolv_conf_buffer == NULL) {
+        PAL_WARN("Warning: Failed to malloc for /etc/resolv.conf buffer");
+        return NULL;
+    }
+    fread(resolv_conf_buffer, 1, fsize, fp);
+    resolv_conf_buffer[fsize] = 0;
+    fclose(fp);
+    return resolv_conf_buffer;
+}
--- a/test/resolv_conf/main.c
+++ b/test/resolv_conf/main.c
@@ -1,0 +1,48 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "test.h"
+
+// ============================================================================
+// Helper functions
+// ============================================================================
+char *read_resolv_conf(void) {
+    FILE *fp = fopen("/etc/resolv.conf", "rb");
+    fseek(fp, 0, SEEK_END);
+    long fsize = ftell(fp);
+    fseek(fp, 0, SEEK_SET);
+    char *resolv_conf_buffer = malloc(fsize + 1);
+    if (resolv_conf_buffer == NULL) {
+        printf("ERROR: Failed to malloc for /etc/resolv.conf buffer");
+        return NULL;
+    }
+    fread(resolv_conf_buffer, 1, fsize, fp);
+    fclose(fp);
+    return resolv_conf_buffer;
+}
+
+// ============================================================================
+// Test cases for resolv.conf file
+// ============================================================================
+
+int test_resolv_conf() {
+    char *buffer = read_resolv_conf();
+    if (buffer == NULL) {
+        THROW_ERROR("failed to read resolv.conf");
+    }
+    printf("%s", buffer);
+    free(buffer);
+    buffer = NULL;
+    return 0;
+}
+
+static test_case_t test_cases[] = {
+    TEST_CASE(test_resolv_conf),
+};
+
+// ============================================================================
+// Test suite main
+// ============================================================================
+
+int main() {
+    return test_suite_run(test_cases, ARRAY_SIZE(test_cases));
+}
--- a/demos/grpc/run_client_on_occlum.sh
+++ b/demos/grpc/run_client_on_occlum.sh
@@ -24,7 +24,6 @@ then
 fi
 
 mkdir -p image/etc
-cp /etc/resolv.conf image/etc
 cp ../greeter_client image/bin
 cp $INSTALL_DIR/lib/libprotobuf.so.3.10.0.0 image/lib
 cp $INSTALL_DIR/lib/libcares.so.2 image/lib
--- a/demos/grpc/run_server_on_occlum.sh
+++ b/demos/grpc/run_server_on_occlum.sh
@@ -23,7 +23,6 @@ then
 fi
 
 mkdir -p image/etc
-cp /etc/resolv.conf image/etc
 cp ../greeter_server image/bin
 cp $INSTALL_DIR/lib/libprotobuf.so.3.10.0.0 image/lib
 cp $INSTALL_DIR/lib/libcares.so.2 image/lib
--- a/deps/resolv-conf.patch
+++ b/deps/resolv-conf.patch
@@ -1,0 +1,80 @@
+From 9faee3943a31e359e76c5dfce23918c35fd8f317 Mon Sep 17 00:00:00 2001
+From: ClawSeven <zehuan97@gmail.com>
+Date: Fri, 9 Jul 2021 13:28:42 +0800
+Subject: [PATCH] Transplant resolv-conf into sgx-world
+
+---
+ Cargo.toml     | 3 ++-
+ src/config.rs  | 2 ++
+ src/grammar.rs | 1 +
+ src/ip.rs      | 1 +
+ src/lib.rs     | 3 +++
+ 5 files changed, 9 insertions(+), 1 deletion(-)
+
+diff --git a/Cargo.toml b/Cargo.toml
+index 0022252..617e319 100644
+--- a/Cargo.toml
++++ b/Cargo.toml
+@@ -14,7 +14,8 @@ version = "0.7.0"
+ authors = ["paul@colomiets.name"]
+ 
+ [dependencies]
+-quick-error = "1.0.0"
++quick-error = {  git = "https://github.com/mesalock-linux/quick-error-sgx.git"  }
++sgx_tstd = { git = "https://github.com/apache/teaclave-sgx-sdk.git" }
+ hostname = { version = "^0.3", optional = true }
+ 
+ [features]
+diff --git a/src/config.rs b/src/config.rs
+index b6b6460..8389064 100644
+--- a/src/config.rs
++++ b/src/config.rs
+@@ -3,6 +3,8 @@ use std::iter::{IntoIterator, Iterator};
+ use std::slice::Iter;
+ use {grammar, Network, ParseError, ScopedIp};
+ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
++use std::vec::Vec;
++use std::string::String;
+ 
+ const NAMESERVER_LIMIT:usize = 3;
+ const SEARCH_LIMIT:usize = 6;
+diff --git a/src/grammar.rs b/src/grammar.rs
+index 9a274a8..be85081 100644
+--- a/src/grammar.rs
++++ b/src/grammar.rs
+@@ -1,5 +1,6 @@
+ use std::net::{Ipv4Addr, Ipv6Addr};
+ use std::str::{Utf8Error, from_utf8};
++use std::string::ToString;
+ 
+ use {AddrParseError, Config, Network, Lookup, Family};
+ 
+diff --git a/src/ip.rs b/src/ip.rs
+index c9cfa59..4b4bca2 100644
+--- a/src/ip.rs
++++ b/src/ip.rs
+@@ -2,6 +2,7 @@ use std::error::Error;
+ use std::fmt;
+ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+ use std::str::FromStr;
++use std::string::{String, ToString};
+ 
+ /// A network, that is an IP address and a mask
+ #[derive(Clone, Debug, PartialEq, Eq)]
+diff --git a/src/lib.rs b/src/lib.rs
+index a55a9d7..094a35a 100644
+--- a/src/lib.rs
++++ b/src/lib.rs
+@@ -91,6 +91,9 @@
+ #![warn(missing_debug_implementations)]
+ #![warn(missing_docs)]
+ 
++#![no_std]
++#[macro_use]
++extern crate sgx_tstd as std;
+ #[macro_use]
+ extern crate quick_error;
+ #[cfg(feature = "system")]
+-- 
+2.17.1
+
--- a/src/Enclave.edl
+++ b/src/Enclave.edl
@@ -42,7 +42,7 @@ enclave {
          *      EEXIST - The LibOS has already been initialized.
          *      EINVAL - The value of an argument are invalid.
          */
-        public int occlum_ecall_init([in, string] const char* log_level, [in, string] const char* instance_dir);
+        public int occlum_ecall_init([in, string] const char* log_level, [in, string] const char* instance_dir, [in, string] const char* resolv_conf_ptr);
 
         /*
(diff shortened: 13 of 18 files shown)


src/libos/src/entry.rs
```diff
@@ -87,6 +89,28 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
         unsafe { backtrace::enable_backtrace(&ENCLAVE_PATH, PrintFormat::Short) };
     });
 
+    let resolv_conf_bytes = unsafe {
+        assert!(!resolv_conf_ptr.is_null());
+        CStr::from_ptr(resolv_conf_ptr).to_bytes()
+    };
+    
+    let cfg = match resolv_conf::Config::parse(resolv_conf_bytes) {
```
guzongmin, line 97: Should copy the string into enclave before using it.

src/libos/src/fs/fs_ops/mount.rs
```diff
@@ -27,5 +28,13 @@ pub fn do_mount_rootfs(
...
         *ENTRY_POINTS.write().unwrap() = user_config.entry_points.to_owned();
     });
+    let fs_view = FsView::new();
+    let resolv_conf_file = match fs_view.open_file("/etc/resolv.conf", AccessMode::O_RDWR as u32 | CreationFlags::O_CREAT.bits(), 0o666){
+        Err(e) => {
+            return_errno!(EINVAL, "failed to open /etc/resolv.conf in enclave");
+        }
+        Ok(file) => file,
+    };
+    resolv_conf_file.write(&*RESOLV_CONF_BYTES.read().unwrap());
```
guzongmin, line 38: Should RESOLV_CONF_BYTES be initialized?

guzongmin (commented):
Needs a test case to test this new feature

src/libos/src/entry.rs
```diff
@@ -87,6 +89,28 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
         unsafe { backtrace::enable_backtrace(&ENCLAVE_PATH, PrintFormat::Short) };
     });
 
+    let resolv_conf_bytes = unsafe {
```
guzongmin, line 92: Can we use from_bytes_with_nul to avoid using unsafe function?

src/libos/src/fs/fs_ops/mount.rs
```diff
@@ -27,5 +28,13 @@ pub fn do_mount_rootfs(
         *root_inode = new_root_inode;
         *ENTRY_POINTS.write().unwrap() = user_config.entry_points.to_owned();
     });
+    let fs_view = FsView::new();
+    let resolv_conf_file = match fs_view.open_file("/etc/resolv.conf", AccessMode::O_RDWR as u32 | CreationFlags::O_CREAT.bits(), 0o666){
```
guzongmin, line 32: Should check whether the file is already there? If the user copied the file, then we should do nothing.

src/libos/src/fs/fs_ops/mount.rs
ClawSeven: Yes, since we need to use RESOLV_CONF_BYTES as a mutable global variable in mount.rs,  RESOLV_CONF_BYTES requires lazy initialization.

src/libos/src/entry.rs
ClawSeven: I can try it

src/libos/src/fs/fs_ops/mount.rs
ClawSeven: Ok， I will consider this situation.

src/libos/src/entry.rs
ClawSeven: Actually, the "CStr::from_ptr" function has copied the pointer and transfers it into slice using slice::from_raw_parts.

src/libos/src/fs/fs_ops/mount.rs
ClawSeven: Done

src/libos/src/entry.rs
ClawSeven: Since the parsed parameter is a pointer, we can't use from_bytes_with_nul. The from_ptr method is unsafe for a number of reasons:

> There is no guarantee to the validity of ptr.
> The returned lifetime is not guaranteed to be the actual lifetime of ptr.
> There is no guarantee that the memory pointed to by ptr contains a valid nul terminator byte at the end of the string.
> It is not guaranteed that the memory pointed by ptr won’t change before the CStr has been destroyed.

src/pal/src/pal_init_read.c
```diff
@@ -0,0 +1,14 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+char *read_resolv_conf(void) {
+    FILE *fp = fopen("/etc/resolv.conf", "rb");
+    fseek(fp, 0, SEEK_END);
+    long fsize = ftell(fp);
+    fseek(fp, 0, SEEK_SET);
+    char *resolv_conf_buffer = malloc(fsize + 1);
```
guzongmin, line 9: Memory leak, no-one would free this buffer.

src/libos/src/fs/fs_ops/mount.rs
ClawSeven: Done

test/resolv_conf/main.c
```diff
@@ -0,0 +1,39 @@
...
+
+// ============================================================================
+// Helper functions
+// ============================================================================
+char *read_resolv_conf(void) {
+    FILE *fp = fopen("/etc/resolv.conf", "rb");
+    fseek(fp, 0, SEEK_END);
+    long fsize = ftell(fp);
+    fseek(fp, 0, SEEK_SET);
+    char *resolv_conf_buffer = malloc(fsize + 1);
```
guzongmin, line 13: Check the buffer is NULL or not

test/resolv_conf/main.c
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// ============================================================================
+// Test cases for resolv.conf file
+// ============================================================================
+
+int test_resolv_conf() {
+    char *buffer = read_resolv_conf();
+    printf("%s", buffer);
+    return 0;
```
guzongmin, line 26: free the buffer

src/pal/src/pal_init_read.c
ClawSeven: Done

test/resolv_conf/main.c
ClawSeven: Done

test/resolv_conf/main.c
ClawSeven: Done

.gitmodules
```diff
@@ -22,3 +22,12 @@
...
 	url = https://github.com/agerasev/ringbuf.git
+[submodule "deps/resolv-conf"]
+	path = deps/resolv-conf
+	url = https://github.com/tailhook/resolv-conf.git
+[submodule "deps/quick-error"]
+	path = deps/quick-error
+	url = https://github.com/tailhook/quick-error.git
+[submodule "deps/no-std-net"]
+	path = deps/no-std-net
+	url = https://github.com/dunmatt/no-std-net.git
```
tatetian, line 33: Personally, I really found these Git submodules annoying. I would really like to have as few of them as possible. What's more annoying is the patch files for them. Hard to read, hard to maintain.

I am ok with putting `resolv-conf` under the `deps` directory, but not with `quick-error` and `no-std-net` crates. They are only indirect dependencies (Occlum does not depend only them, `resolv-conf` does). But their appearance in `deps` seem to suggest they are direct dependencies. If we keep adding submodules into `deps` this way, the complexity of the directory would quickly get out of control. And the `no-std-net` crate looks especially dubious as its code base is very immature. Thus, depending on it does not seem a good idea to me.

I looked through the patch files. I cannot see why `resolv-conf` cannot be ported with Rust SGX SDK. By using Rust SGX SDK, we can at least get away with `no-std-net` crate (plus the `core-error` crate, which is even worse than `no-std-net` as a dependency).

src/libos/src/entry.rs
```diff
@@ -23,6 +24,7 @@ lazy_static! {
     static ref HAS_INIT: AtomicBool = AtomicBool::new(false);
     pub static ref ENTRY_POINTS: RwLock<Vec<PathBuf>> =
         RwLock::new(config::LIBOS_CONFIG.entry_points.clone());
+    pub static ref RESOLV_CONF_BYTES: RwLock<Vec<u8>> = RwLock::new(Vec::new());
```
tatetian, line 27: Using `String` seems better than `Vec<u8>`. First, the content of any well-formatted `resolve.conf` is a string, right? Second, `String` is much more convenient than `Vec<u8>` for debugging purposes.

src/libos/src/lib.rs
```diff
@@ -54,6 +54,9 @@ extern crate serde_json;
 #[macro_use]
 extern crate memoffset;
```
tatetian, line 56: Remove the blanket line here so that all crate declarations are together.

src/libos/src/entry.rs
```diff
@@ -87,6 +93,39 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
...
+    // Read resolv.conf file from host
+    let resolv_conf_bytes = unsafe {
+        assert!(!resolv_conf_ptr.is_null());
+        CStr::from_ptr(resolv_conf_ptr).to_bytes()
+    };
+
+    // Parse and inspect resolv.conf file
+    match resolv_conf::Config::parse(resolv_conf_bytes) {
+        Err(e) => {
+            eprintln!("invalid host /etc/resolv.conf: {}", e);
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -87,6 +93,39 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
...
     });
 
+    // Read resolv.conf file from host
+    let resolv_conf_bytes = unsafe {
+        assert!(!resolv_conf_ptr.is_null());
+        CStr::from_ptr(resolv_conf_ptr).to_bytes()
+    };
+
+    // Parse and inspect resolv.conf file
+    match resolv_conf::Config::parse(resolv_conf_bytes) {
```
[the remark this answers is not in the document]
ClawSeven: Here is a logical problem, if the copied resolv.conf is malformatted, we think it it appropriate to make occlum crash, so we decide to drop the malformatted resolv.conf rather than return when parsing error appear.

src/pal/src/pal_init_read.c
```diff
@@ -0,0 +1,22 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "pal_log.h"
+
+char *read_resolv_conf(void) {
+    FILE *fp = fopen("/etc/resolv.conf", "rb");
+
+    if (fp == NULL) {
+        PAL_WARN("Warning: Failed to open /etc/resolv.conf file");
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -87,6 +93,39 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
         unsafe { backtrace::enable_backtrace(&ENCLAVE_PATH, PrintFormat::Short) };
     });
 
+    // Read resolv.conf file from host
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -87,6 +93,39 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
...
+                AccessMode::O_RDWR as u32 | CreationFlags::O_CREAT.bits(),
+                0o666,
+            ) {
+                Err(e) => {
+                    eprintln!("failed to open /etc/resolv.conf in enclave: {}", e);
+                    return ecall_errno!(EINVAL);
+                }
+                Ok(file) => file,
+            };
+            resolv_conf_file.write(resolv_conf_bytes);
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/fs/fs_ops/mount.rs
```diff
@@ -27,5 +30,26 @@ pub fn do_mount_rootfs(
         *root_inode = new_root_inode;
         *ENTRY_POINTS.write().unwrap() = user_config.entry_points.to_owned();
     });
+
+    let resolv_conf_path = String::from("/etc/resolv.conf");
```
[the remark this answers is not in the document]
ClawSeven: Done

src/pal/src/pal_init_read.c
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    if (fp != NULL) {
+        fseek(fp, 0, SEEK_END);
+        long fsize = ftell(fp);
+        fseek(fp, 0, SEEK_SET);
+        char *resolv_conf_buffer = malloc(fsize + 1);
```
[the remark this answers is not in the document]
ClawSeven: Done

src/pal/src/pal_init_read.c
```diff
@@ -0,0 +1,22 @@
...
+        fread(resolv_conf_buffer, 1, fsize, fp);
+        fclose(fp);
+        return resolv_conf_buffer;
```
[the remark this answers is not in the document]
ClawSeven: Done

src/pal/src/pal_init_read.h
```diff
@@ -0,0 +1,4 @@
+#ifndef __PAL_INIT_READ_H__
+#define __PAL_INIT_READ_H__
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/lib.rs
ClawSeven: Done

src/libos/src/entry.rs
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -87,6 +93,39 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
...
+
+            // Prepare resolv.conf file for occlum image fs
+            let mut resolv_conf_write = RESOLV_CONF_BYTES.write().unwrap();
+            resolv_conf_write.append(&mut resolv_conf_bytes.to_vec());
```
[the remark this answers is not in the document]
ClawSeven: Done

.gitmodules
ClawSeven: Done

src/libos/src/entry.rs
ClawSeven: Done

src/libos/src/fs/fs_ops/mount.rs
```diff
@@ -27,5 +30,26 @@ pub fn do_mount_rootfs(
...
+        Err(e) if e.errno() == ENOENT => {
+            let resolv_conf_file = fs_view.open_file(
+                RESOLV_CONF_PATH,
+                AccessMode::O_RDWR as u32 | CreationFlags::O_CREAT.bits(),
+                0o666,
+            )?;
+            resolv_conf_file.write(&*RESOLV_CONF_BYTES.read().unwrap().as_bytes());
+        }
+        Err(e) => return Err(e),
+        _ => (),
```
tatetian, line 52: I think we need to give more thought about this case of existing `/etc/resolv.conf` file.

In my opinion, it would be more robust to update the `/etc/resolv.conf` file on SEFS every time the LibOS starts. This way, if the network configuration of the host changes, the LibOS will use the latest configuration on startup.

With that said, I think the current design decision of not overriding existing `/etc/resolv.conf` file also makes sense in some use cases. But if whether overriding existing `/etc/resolv.conf` file is not a configurable policy (as least for now it is not), I would vote for the former approach, which is more robust in face of changing network configurations on host.

src/pal/src/pal_load_resolv_conf.c
```diff
@@ -0,0 +1,21 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "pal_log.h"
+
+char *read_resolv_conf(void) {
```
tatetian, line 5: Readability tips: do not give one concept multiple names. Since we already have a module named `pal_load_resolv_conf`, it only makes sense to name the API `pal_load_resolv_conf` (Yes, with a prefix of `pal_` to be consistent with all other module level APIs).

src/pal/src/pal_load_resolv_conf.c
```diff
@@ -0,0 +1,21 @@
...
+    FILE *fp = fopen("/etc/resolv.conf", "rb");
+
+    if (fp == NULL) {
+        PAL_WARN("Warning: Failed to open /etc/resolv.conf file");
+        return NULL;
+    } else {
+        fseek(fp, 0, SEEK_END);
+        long fsize = ftell(fp);
+        fseek(fp, 0, SEEK_SET);
+        char *resolv_conf_buffer = malloc(fsize + 1);
```
tatetian, line 15: Should check for null.

tatetian (commented):
I see a lot of improvements over the previous version. Especially, reducing the number of crates in `deps` to just one. Good job.

There are still some minor issues. See my inline comments.

Makefile
```diff
@@ -35,6 +35,7 @@ submodule: githooks
 	@# Try to apply the patches. If failed, check if the patches are already applied
 	cd deps/serde-json-sgx && git apply ../serde-json-sgx.patch >/dev/null 2>&1 || git apply ../serde-json-sgx.patch -R --check
 	cd deps/ringbuf && git apply ../ringbuf.patch >/dev/null 2>&1 || git apply ../ringbuf.patch -R --check
+	cd deps/resolv-conf && git checkout c075281 && git apply ../resolv-conf.patch >/dev/null 2>&1 || git apply ../resolv-conf.patch -R --check
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -240,6 +252,32 @@ fn parse_arguments(
     Ok((path_buf, args, env_merged, host_stdio_fds))
 }
 
+fn parse_resolv_conf(resolv_conf_ptr: *const c_char) -> Result<()> {
+    // Read resolv.conf file from host
+    let resolv_conf_bytes = unsafe {
+        assert!(!resolv_conf_ptr.is_null());
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -23,6 +25,7 @@ lazy_static! {
     static ref HAS_INIT: AtomicBool = AtomicBool::new(false);
     pub static ref ENTRY_POINTS: RwLock<Vec<PathBuf>> =
         RwLock::new(config::LIBOS_CONFIG.entry_points.clone());
+    pub static ref RESOLV_CONF_BYTES: RwLock<String> = RwLock::new(String::new());
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -87,6 +94,11 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
         unsafe { backtrace::enable_backtrace(&ENCLAVE_PATH, PrintFormat::Short) };
     });
 
+    // Parse host resolv.conf file
+    if let Err(e) = parse_resolv_conf(resolv_conf_ptr) {
+        eprintln!("failed to parse host /etc/resolv.conf: {}", e);
+    };
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -240,6 +252,32 @@ fn parse_arguments(
...
+    let resolv_conf_file = fs_view.open_file(
+        "/etc/resolv.conf",
+        AccessMode::O_RDWR as u32 | CreationFlags::O_CREAT.bits(),
+        0o666,
+    )?;
+    resolv_conf_file.write(resolv_conf_bytes)?;
+
+    // Prepare resolv.conf file for occlum mounted fs
+    let mut resolv_conf_write = RESOLV_CONF_BYTES.write()?;
+    resolv_conf_write.push_str(str::from_utf8(resolv_conf_bytes).unwrap());
```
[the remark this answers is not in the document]
ClawSeven: Done

src/libos/src/fs/fs_ops/mount.rs
ClawSeven: Done, I had discussed that /etc/resolv.conf may be multi-times copied with Zongmin, we thought overriding the older /etc/resolv.conf is more reasonalble. Good advice!

src/pal/src/pal_load_resolv_conf.c
ClawSeven: Done

src/pal/src/pal_load_resolv_conf.c
ClawSeven: Done

src/libos/src/entry.rs
```diff
@@ -240,6 +252,36 @@ fn parse_arguments(
...
 
+fn parse_resolv_conf(resolv_conf_ptr: *const c_char) -> Result<()> {
+    // Read resolv.conf file from host
+    let resolv_conf_bytes = unsafe { CStr::from_ptr(resolv_conf_ptr).to_bytes() };
+    let resolv_conf_str =
+        str::from_utf8(resolv_conf_bytes).map_err(|_| errno!(ENOENT, "invalid symlink content"))?;
+    // if let resolv_conf_str = match str::from_utf8(resolv_conf_bytes) {
+    //     Err(e) => e,
+    //     Ok(str) => str
+    // }
```
tatetian, line 263: Should these commented lines be removed?

src/libos/src/entry.rs
```diff
@@ -240,6 +252,36 @@ fn parse_arguments(
     Ok((path_buf, args, env_merged, host_stdio_fds))
 }
 
+fn parse_resolv_conf(resolv_conf_ptr: *const c_char) -> Result<()> {
+    // Read resolv.conf file from host
+    let resolv_conf_bytes = unsafe { CStr::from_ptr(resolv_conf_ptr).to_bytes() };
+    let resolv_conf_str =
+        str::from_utf8(resolv_conf_bytes).map_err(|_| errno!(ENOENT, "invalid symlink content"))?;
```
tatetian, line 259: `EINVAL` is more appropriate. And why is the error message `"invalid symlink content"`?

src/libos/src/fs/fs_ops/mount.rs
```diff
@@ -27,5 +30,21 @@ pub fn do_mount_rootfs(
...
+fn write_resolv_conf() -> Result<()> {
+    const RESOLV_CONF_PATH: &'static str = "/etc/resolv.conf";
+    let fs_view = FsView::new();
+    // overwrite /etc/resolv.conf if existed
+    let resolv_conf_file = fs_view.open_file(
+        RESOLV_CONF_PATH,
+        AccessMode::O_RDWR as u32 | CreationFlags::O_CREAT.bits() | CreationFlags::O_TRUNC.bits(),
+        0o666,
+    )?;
+    resolv_conf_file.write(&*RESOLV_CONF_STR.read().unwrap().as_bytes());
```
tatetian, line 48: I think `&*` is not necessary, is it?

src/libos/src/lib.rs
```diff
@@ -53,6 +53,8 @@ extern crate serde;
 extern crate serde_json;
 #[macro_use]
 extern crate memoffset;
+#[macro_use]
```
tatetian, line 56: I didn't notice that we need any macros from the resolve_conf crate, do we?

src/pal/src/pal_load_resolv_conf.c
```diff
@@ -0,0 +1,25 @@
...
+#include <stdlib.h>
+#include "pal_log.h"
+
+char *pal_load_resolv_conf(void) {
+    FILE *fp = fopen("/etc/resolv.conf", "rb");
+
+    if (fp == NULL) {
+        PAL_WARN("Warning: Failed to open /etc/resolv.conf file");
+        return NULL;
+    } else {
```
tatetian, line 11: Style note. We don't need the `else` part. Simply removing the `else` would make the code more readable as the indentation level is reduced.

tatetian (commented):
We are very close to complete this PR. There are some minor issues. See the inline comments. And don't forget to clean up the Git history by squashing all commits into one.

src/libos/src/entry.rs
ClawSeven: Done

src/libos/src/entry.rs
ClawSeven: Done

src/libos/src/fs/fs_ops/mount.rs
ClawSeven: Done

src/pal/src/pal_load_resolv_conf.c
ClawSeven: Done

src/libos/src/lib.rs
ClawSeven: Done, we don't use any macros from resolv_conf crate.

src/libos/src/entry.rs
```diff
@@ -240,6 +252,32 @@ fn parse_arguments(
...
+    let resolv_conf_bytes = unsafe { CStr::from_ptr(resolv_conf_ptr).to_bytes() };
+    let resolv_conf_str = str::from_utf8(resolv_conf_bytes)
+        .map_err(|_| errno!(EINVAL, "/etc/resolv.conf contains non UTF-8 characters"))?;
+
+    // Parse and inspect resolv.conf file
+    if let Err(_) = resolv_conf::Config::parse(resolv_conf_bytes) {
+        return_errno!(EINVAL, "malformated host /etc/resolv.conf");
+    }
+
+    let fs_view = FsView::new();
```
guzongmin, line 266: This function not only parse the resolv.conf file but also copy it to a local file. Should be a separate function  to do the copy.

src/libos/src/entry.rs
```diff
@@ -23,6 +25,7 @@ lazy_static! {
     static ref HAS_INIT: AtomicBool = AtomicBool::new(false);
     pub static ref ENTRY_POINTS: RwLock<Vec<PathBuf>> =
         RwLock::new(config::LIBOS_CONFIG.entry_points.clone());
+    pub static ref RESOLV_CONF_STR: RwLock<String> = RwLock::new(String::new());
```
liqinggd, line 28: Suggestion:
```
pub static ref RESOLV_CONF_STR: RwLock<Option<String>> = RwLock::new(None);
```
Then you can initialize the `RESOLV_CONF_STR` with a valid string. After using the `RESOLV_CONF_STR`, you can just set it to `None` to make the buffer of string released.

tatetian: This make senses. The size of a resolve.conf file could be large. Releasing the `String` after it is no longer useful is good.

src/libos/src/entry.rs
tatetian: Good point.

tatetian:
@ClawSeven Let's fix the latest few issues and get this PR done 🔥

src/libos/src/entry.rs
```diff
@@ -240,6 +265,31 @@ fn parse_arguments(
...
+    let resolv_conf_bytes = unsafe { CStr::from_ptr(resolv_conf_ptr).to_bytes() };
+    let resolv_conf_str = str::from_utf8(resolv_conf_bytes)
+        .map_err(|_| errno!(EINVAL, "/etc/resolv.conf contains non UTF-8 characters"))?;
+
+    // Parse and inspect resolv.conf file
+    if let Err(_) = resolv_conf::Config::parse(resolv_conf_bytes) {
+        return_errno!(EINVAL, "malformated host /etc/resolv.conf");
+    }
+
+    // let fs_view = FsView::new();
```
guzongmin, line 279: Please remove the code instead of leave them as comments.

src/libos/src/util/resolv_conf_util.rs
```diff
@@ -0,0 +1,21 @@
...
+pub fn write_resolv_conf() -> Result<()> {
+    const RESOLV_CONF_PATH: &'static str = "/etc/resolv.conf";
+    let fs_view = FsView::new();
+    // overwrite /etc/resolv.conf if existed
+    let resolv_conf_file = fs_view.open_file(
+        RESOLV_CONF_PATH,
+        AccessMode::O_RDWR as u32 | CreationFlags::O_CREAT.bits() | CreationFlags::O_TRUNC.bits(),
+        0o666,
+    )?;
+    let resolv_conf_str = RESOLV_CONF_STR.read().unwrap().clone();
```
guzongmin, line 13: Since no other thread expects to get the lock, does no need "clone".

src/libos/src/entry.rs
```diff
@@ -87,6 +95,23 @@ pub extern "C" fn occlum_ecall_init(log_level: *const c_char, instance_dir: *con
...
     });
 
+    // Parse host resolv.conf file
+    if let Err(e) = parse_resolv_conf(resolv_conf_ptr) {
+        eprintln!("failed to parse host /etc/resolv.conf: {}", e);
+    }
+
+    match parse_resolv_conf(resolv_conf_ptr) {
+        Err(e) => {
+            eprintln!("failed to parse host /etc/resolv.conf: {}", e);
```
guzongmin, line 105: Use "debug!" instead of "eprintln!"

ClawSeven: Done, use "error" to replace "eprintln".

src/libos/src/util/resolv_conf_util.rs
ClawSeven: Done

src/libos/src/entry.rs
ClawSeven: Done

src/libos/src/util/mod.rs
```diff
@@ -5,5 +5,6 @@ pub mod log;
 pub mod mem_util;
 pub mod mpx_util;
 pub mod random;
+pub mod resolv_conf_util;
```
tatetian, line 8: Although I wanted to point out that encapsulating all resolve_conf related stuff into a separate module is a good idea, I kept it to myself as I thought the original implementation was acceptable. Now that you have added the `resolv_conf_util` module,  I believe you also see the benefit of reconstructing all resolve_conf related code as a module.  **If a resolve_conf module is to be added, then it should be done properly: give it well-designed public APIs and hide all implementation details.** A good example is `crate::util::log`, which is initialized at the startup of the enclave (as the resolve_conf module would do) and used later by other modules (as the resolve_conf module would do, too).

ClawSeven: Done

guzongmin (approved):
LGTM

PR #485, opened 2021-06-23
Fix readme doc for new dcap or in-tree kernel driver

jessehui:
Reference:
[1] https://github.com/intel/linux-sgx/blob/f1fcf9175d58d28bbee576438d9b68ecf93f5e4e/psw/urts/linux/edmm_utility.cpp#L89
[2] https://www.kernel.org/doc/html/v5.11/x86/sgx.html#enclave-build-functions


guzongmin (approved):
LGTM

PR #490, opened 2021-06-27
Support to run Flink 1.10 in Occlum

SteNicholas:


guzongmin:
Please update the action file https://github.com/occlum/occlum/blob/master/.github/workflows/demo_test.yml to add this demo into auto-test list.

demos/flink/flink-1.10/run_flink_on_occlum_glibc.sh
```diff
@@ -0,0 +1,153 @@
...
+    build_flink_and_fish
+    echo -e "${BLUE}occlum run Flink task${NC}"
+    echo -e "${BLUE}logfile=$log${NC}"
+    occlum run /bin/bin/flink run /bin/examples/streaming/WordCount.jar
+}
+
+id=$([ -f "$pid" ] && echo $(wc -l < "$pid") || echo "0")
+
+arg=$1
+case "$arg" in
```
guzongmin, line 140: If adding one more parameter "environment (in occlum or native)" to help user select how to run the flink modules, it would be better. For example, someone only wants to run tm in occlum, then the use could run tm with parameter "occlum" and run jm and task with parameter "native".

PR #510, opened 2021-07-09
Add occlum-trace tool

StevenJiang1110:
This PR adds an `occlum-trace` tool to cope with occlum's trace log (`OCCLUM_LOG_LEVEL=trace`).
The aim of this tool can be seen in the `README.md`.

### DIFF
--- a/.github/workflows/hw_mode_test.yml
+++ b/.github/workflows/hw_mode_test.yml
@@ -205,7 +205,7 @@ jobs:
 
     - name: Go server set up and run
       run: docker exec $language_support_test bash -c "export GO111MODULE=on && export GOPROXY=https://goproxy.cn;
-            cd /root/occlum/demos/golang/web_server && occlum-go get -u -v github.com/gin-gonic/gin;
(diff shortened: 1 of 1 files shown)


tools/occlum-trace/src/main.rs
```diff
@@ -0,0 +1,768 @@
...
+        Regex::new(OPEN_AT_RELATIVE_PATTERN).unwrap()
+    };
+    static ref FILENAME_REGEX: Regex = {
+        Regex::new(FILENAME_PATTERN).unwrap()
+    };
+}
+
+
+#[derive(Debug, Clone)]
+struct TraceOption {
```
guzongmin, line 73: Occlum is using clap in other tools. Maybe we could use the same crate in this tool.

guzongmin (commented):
Is occlum-pretty-trace a better name for this tool?

tools/occlum-trace/src/main.rs
StevenJiang1110: Fixed.

PR #512, opened 2021-07-13
Add support for OpenVINO 2021.3

a3213105:
1. Replace the original TBB with OneTBB v2020.3
2. Replace the 2019_R3 with DLDT 2021.3
3. Add some dependent lib to the occlum image

### DIFF
--- a/demos/openvino/run_benchmark_on_occlum.sh
+++ b/demos/openvino/run_benchmark_on_occlum.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 benchmark=benchmark_app
-inference_bin=openvino_src/inference-engine/bin/intel64/Release
+inference_bin=openvino_src/bin/intel64/Release
 occlum_lib=/usr/local/occlum/x86_64-linux-musl/lib
 set -e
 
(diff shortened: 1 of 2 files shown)


demos/openvino/run_benchmark_on_occlum.sh
```diff
@@ -9,25 +9,39 @@ rm -rf occlum_instance
 mkdir occlum_instance
 cd occlum_instance
 occlum init
+cpu_cc=`cat /proc/cpuinfo | grep processor | wc -l`
+#new_json="$(jq '.resource_limits.user_space_size = "4GB" |
```
guzongmin, line 13: If these configs are not needed, can we remove them?

guzongmin (commented):
It is so great that the latest OpenVINO is able to work in Occlum without any change. I very appreciate you to update the OpenVINO to the version 2021.3. 
Only one small comment: Can you please update the commit message from "add support for OpenVINO 2021.3" to "Add support for OpenVINO 2021.3"?

demos/openvino/run_benchmark_on_occlum.sh
a3213105: yes, but for some multi-core/multi-socket servers , the default configs does not work properly， the commented out configs is for that， it can make users  to test more efficient

guzongmin (approved):
LGTM

PR #514, opened 2021-07-13
Remove the cpu online file for glibc

liqinggd:
Checkout here for the implementation of `__get_nprocs `,
https://github.com/bminor/glibc/blob/glibc-2.27/sysdeps/unix/sysv/linux/getsysstats.c#L214

Since `/proc/cpuinfo` is supported in Occlum, we can remove the online file from the image.


guzongmin (approved):
LGTM

PR #534, opened 2021-07-19
Fix the return value of sched_getaffinity.

ShuochengWang:
The return value of the raw sched_getaffinity syscall must be an integer multiple of sizeof(long) in Linux.

Linux code: 
```
unsigned int retlen = min(len, cpumask_size());

static inline unsigned int cpumask_size(void)
{
	return BITS_TO_LONGS(nr_cpumask_bits) * sizeof(long);
}
```


guzongmin:
Please help to confirm if the syscall parameter of buffer size is not 8*x, what is the correct return value?

guzongmin (approved):
LGTM

PR #548, opened 2021-07-21
Fix some process VM error when enabling GCC 10

jessehui:
(1) Fix uninitialized elf memory for loadable segment

(2) Fix an error when calculating elf memory usage
VMLayout was mistakenly used to calculate the memory usage. This
commit is to fix this and separate VMLayout "add" and "extend"
methods.

(3) Don't allocate extra memory if already aligned

### DIFF
--- a/src/libos/src/vm/process_vm.rs
+++ b/src/libos/src/vm/process_vm.rs
@@ -101,7 +101,7 @@ impl<'a, 'b> ProcessVMBuilder<'a, 'b> {
         let process_layout = elf_layouts.iter().chain(other_layouts.iter()).fold(
             VMLayout::new_empty(),
             |mut process_layout, sub_layout| {
-                process_layout.extend(&sub_layout);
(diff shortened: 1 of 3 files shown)


src/libos/src/vm/process_vm.rs
```diff
@@ -248,7 +248,7 @@ impl<'a, 'b> ProcessVMBuilder<'a, 'b> {
                 );
 
                 // Set the remaining part to zero based on alignment
-                empty_end_offset = align_up(mem_start_offset + file_size, alignment);
+                empty_end_offset = align_up(mem_start_offset + max(file_size, mem_size), alignment);
```
guzongmin, line 247: file_size should always smaller than mem_size. We can add assertion to confirm it.

jessehui: Done. Thanks for your review.

guzongmin (approved):
LGTM

Issue #542, opened 2021-07-19
[RFC]Manage the Occlum image with manifest file


guzongmin:
- Feature Name: Occlum image configuration tool set 
- Start Date: 2021-07-19

# Summary & Motivation
[summary]: #summary
All the files in the Occlum image are trusted. So it is very important for the user to know what are installed in the image.
Occlum user needs a manifest of the image folder. The manifest could be used in several ways. 
+ The user could copy correct files into the image folder by the manifest.
+ The user could review the manifest to confirm the Occlum image is expected.

So we need at least two tools:
+ generate the manifest
+ copy the correct files to image folder with a manifest

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation
## 0. manifest file
A manifest file should:
+ has multiple entries: Each file should have at least one entry in the manifest. Occlum image has a lost of files, so a manifest would have multiple entries.
+ has the ability to include another manifest file: most of the application would link with libc libraries. An easy way for user is to include a libc manifest file.

A manifest entry should:
+ has a source 
+ has a target
+ has a hash value if it is not a folder 

## 1. generate_Occlum_manifest
+ The input could be a file
+ If the input file is an executable file, user could ask the tool to add the dependency files too
+ The input could be a folder. The tool would add all the contents under the folder into manifest file recursively 

## 2. copy_with_manifest
+ for each entry from the manifest, this tool copy the source to target path and check the hash value is correct

## 3. update_manifest
+ update the hash value with the latest binaries

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

The manifest file appends with the tools are used to replace the "cp" commands in the configuration script. The manifest file would clearly show what files are copied into the image as expected. 

# Drawbacks
[drawbacks]: #drawbacks

In the current usage, Occlum works well without this feature. But it only because that Occlum is not used in complex environment.

# Unresolved questions
[unresolved-questions]: #unresolved-questions

If the user use the tool to add an executable file and want to add the dependency files tool. How could the tool know what is the correct target path of the dependency files?

PR #550, opened 2021-07-22
Add tools to manage occlum image with configuration files

StevenJiang1110:
add tools described in #542


tools/occlum-bom/README.md
```diff
@@ -0,0 +1,43 @@
...
+More options of these tools can be seen by `<tool-name> --help`.
+
+## How to install this tool
+
+`cd occlum-bom && make install`
+<font color="#660000"> Note: before use these tool, `rsync` should be installed to copy files (`cp` command can't support symbolic link) </font><br />
+`apt install rsync -y`
+
+## Example For hello_c in occlum docker image (After install this tool):
+`cd demos/hello_c && make`
```
guzongmin, line 39: We'd be better to use this tool in the hello_c demo instead of only add it in the README file.

PR #556, opened 2021-07-26
Add support for /dev/fd

jessehui:


liqinggd:
I think the current implementation is too complicated，you can just implement `read_at` and  the`metadata` method for the INode is enough.

When encounter  the `/dev/fd`, `read_at` method returns the bytes`/proc/self/fd`.

jessehui:
@liqinggd Got it. I will take a look. Thank you!

jessehui:
@liqinggd Thank you for the suggestion! Done.

src/libos/src/fs/dev_fs/dev_fd.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn read_at(&self, offset: usize, buf: &mut [u8]) -> vfs::Result<usize> {
+        let proc_fd = "/proc/self/fd";
+        for (tgt, src) in buf.iter_mut().zip(proc_fd.as_bytes().iter()) {
+            *tgt = *src;
+        }
+        Ok(proc_fd.len())
+    }
+
+    fn write_at(&self, offset: usize, buf: &[u8]) -> vfs::Result<usize> {
+        Err(vfs::FsError::NotFile)
```
liqinggd, line 15: I think the `NotFile` error here is not correct, because symlinks are also allowed to write the new symlink content. In this case, the symlink cannot be modified, we can return 'FsError::PermError'

test/device/main.c
```diff
@@ -143,6 +144,45 @@ int test_dev_shm() {
...
+        THROW_ERROR("calloc failed");
+    }
+    asprintf(&fd_str, "%d", fd);
+    strcat(dev_fd_path, fd_str);
+    int dev_fd = open(dev_fd_path, O_WRONLY, 0666);
+    if (dev_fd < 0) {
+        THROW_ERROR("failed to open %s", dev_fd_path);
+    }
+
+    int ret = write(dev_fd, greetings, strlen(greetings));
```
liqinggd, line 168: use 'len' as the return value of write/read is more correct.

test/device/main.c
```diff
@@ -143,6 +144,45 @@ int test_dev_shm() {
...
+    ret = read(fd, buf, ret);
+    if (ret < 0) {
+        THROW_ERROR("failed to read from %s", file_path);
+    }
+
+    if (strcmp(buf, greetings)) {
+        THROW_ERROR("file content is wrong");
+    }
+
+    free(fd_str);
```
liqinggd, line 183: please the opened fds after use

src/libos/src/fs/dev_fs/dev_fd.rs
jessehui: Done.

test/device/main.c
jessehui: Done.

test/device/main.c
jessehui: Done. Thanks for your review.

liqinggd:
LGTM

guzongmin (approved):
LGTM

PR #558, opened 2021-07-27
Add partial support for '/proc/stat' and '/proc/[pid]/stat'

liqinggd:


src/libos/src/fs/procfs/stat_inode.rs
```diff
@@ -0,0 +1,20 @@
...
+impl StatINode {
+    pub fn new() -> Arc<dyn INode> {
+        Arc::new(File::new(Self))
+    }
+}
+
+impl ProcINode for StatINode {
+    fn generate_data_in_bytes(&self) -> vfs::Result<Vec<u8>> {
+        Ok(fs::read_to_string("/proc/stat")
+            .map_err(|e| e.into_fs_error())?
```
guzongmin, line 17: Can we do any simple check to confirm it is really a /proc/stat file?

liqinggd: I think the check here is not trivial,  you have to check the file if it is in some predefined format, maybe some regex match tools can be used here. BTW, I think the effort is more than just implement a safe one.

guolanting:
@guzongmin , Hi, what is going on about supporting '/proc/stat'.  I just encountered a "/proc/stat" error when using Occlum.

liqinggd:
Update the commit with the newer implementation

src/libos/src/time/up_time.rs
```diff
@@ -2,7 +2,7 @@ use super::{do_clock_gettime, ClockID};
 use std::time::Duration;
 
 lazy_static! {
-    static ref BOOT_TIME_STAMP: Duration = do_clock_gettime(ClockID::CLOCK_MONOTONIC_RAW)
+    static ref BOOT_TIME_STAMP: Duration = do_clock_gettime(ClockID::CLOCK_REALTIME)
```
volcano0dr, line 5: Why change stable monotonic time to wall time？

src/libos/src/fs/procfs/stat.rs
```diff
@@ -0,0 +1,39 @@
...
+            processes.retain(|p| p.status() == crate::process::ProcessStatus::Running);
+            processes.len().to_string()
+        };
+        fill_in_stat(&mut host_stat, "procs_running", &procs_running);
+        Ok(host_stat.into_bytes())
+    }
+}
+
+fn fill_in_stat(stat: &mut String, pat: &str, val: &str) {
+    let start_idx = stat.find(pat).unwrap() + pat.len() + 1;
```
volcano0dr, line 31: If find fails, `unwrap` will generate panic, I think it should return failure instead of panic.

src/libos/src/fs/procfs/stat.rs
```diff
@@ -0,0 +1,39 @@
...
+    let end_idx = stat
+        .chars()
+        .skip(start_idx)
+        .position(|c| c == '\n')
+        .unwrap()
```
volcano0dr, line 36: replace `unwrap` with return error

volcano0dr (approved):
LGTM

PR #563, opened 2021-07-27
Add getdents support for procfs's inode

liqinggd:


src/libos/src/fs/procfs/proc_inode/mod.rs
```diff
@@ -53,8 +54,63 @@ macro_rules! impl_inode_for_file_or_symlink {
...
+            Err(e) => {
+                if *total_written == 0 {
+                    return Err(e);
+                } else {
+                    return Ok(*total_written);
+                }
+            }
+        }
+    };
+}
```
guzongmin, line 92:
```suggestion
macro_rules! write_inode_entry {
    ($ctx:expr, $name:expr, $inode:expr, $total_written:expr) => {
        let ctx = $ctx;
        let name = $name;
        let inode = $inode;
        let total_written = $total_written;
        let ino = inode.metadata()?.inode as u64;
        let type_ = inode.metadata()?.type_;

        write_entry!(ctx, name, ino, type, total_written);
    };
}
```

liqinggd: Good suggestion

src/libos/src/fs/procfs/mod.rs
```diff
@@ -135,13 +141,61 @@ impl DirProcINode for LockedProcRootINode {
...
+        let mut total_written_len = 0;
+        let idx = ctx.pos();
+
+        // Write first two special entries
+        if idx == 0 {
+            let this_inode = file.this.upgrade().unwrap();
+            write_inode_entry!(&mut ctx, ".", &this_inode, &mut total_written_len);
+        }
+        if idx <= 1 {
+            let parent_inode = file.parent.upgrade().unwrap();
```
guzongmin, line 170: "let parent_inode = file.parent.upgrade().unwrap();" only used in this file. What is the reason?

src/libos/src/fs/procfs/pid_inode.rs
```diff
@@ -121,11 +145,53 @@ impl DirProcINode for LockedProcFdDirINode {
...
+        let idx = ctx.pos();
+
+        // Write first two special entries
+        if idx == 0 {
+            let this_inode = file.this.upgrade().unwrap();
+            write_inode_entry!(&mut ctx, ".", &this_inode, &mut total_written_len);
+        }
+        if idx <= 1 {
+            write_inode_entry!(&mut ctx, "..", &file.parent, &mut total_written_len);
+        }
```
guzongmin, line 172: Can we add a function to replace these code in all the procfs files?

src/libos/src/fs/procfs/mod.rs
liqinggd: This is because the parent is the root inode itself,  since the inode itself exists, the pointer can be weak. Maybe we can just remove the parent member of root inode of procfs.

src/libos/src/fs/procfs/pid_inode.rs
liqinggd: I think add a macro is good option

guzongmin (approved):
LGTM

PR #571, opened 2021-08-03
Remove the mount config for tmp dir

### LINKED ISSUE #570: [RFC] Remove the mount config for the "/tmp" directory.
- Feature Name:  Remove the mount config for the "/tmp" directory
- Start Date: 2021-08-03

# Summary
[summary]: #summary

This RFC requests to remove the mount config for the "/tmp" directory.

# Motivation
[motivation]: #motivation

## Temporary files management on Linux

The "/tmp" directory is managed by `systemd-tmpfiles` in many popular Linux Distributions(e.g. CentOS/RHEL/Ubuntu) which use Systemd as their init process(PID = 1).

When system bootup, it executes `systemd-tmpfiles-setup.service` to create and delete files in some folders. It reads configuration files from `/usr/lib/tmpfiles.d/*.conf`, `/run/tmpfiles.d/*.conf`, and `/etc/tmpfiles.d/*.conf`. 

Here is an example of tmp.conf
```
cat /usr/lib/tmpfiles.d/tmp.conf
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.

# See tmpfiles.d(5) for details

# Clear tmp directories separately, to make them easier to override
D /tmp 1777 root root -
#q /var/tmp 1777 root root 30d

# Exclude namespace mountpoints created with PrivateTmp=yes
x /tmp/systemd-private-%b-*
X /tmp/systemd-private-%b-*/tmp
x /var/tmp/systemd-private-%b-*
X /var/tmp/systemd-private-%b-*/tmp

# Remove top-level private temporary directories on each boot
R! /tmp/systemd-private-*
R! /var/tmp/systemd-private-*
```

To ensure that long-running systems do not fill up their disks with stale data,  a systemd timer unit called `systemd-tmpfiles-clean.timer` triggers the `systemd-tmpfiles-clean.service` on a regular interval to clean the temporary files.

## Temporary files management on Docker container

Generally speaking, the writable layer of the container is where all the data will be written in the container if no volume is mounted. It will persist on container restart but will be deleted if the container is deleted. So "/tmp" dir inside the container is persisted until one deletes the container.

One can mount a tmpfs volume which will persist data as long as the container is running (i.e. the data in the volume will be deleted when the container stops), for example:
```
docker run --mount type=tmpfs,destination=/home/tmp_dir some_image
```
A tmpfs mount is temporary and only persisted in the host memory. When the container stops, the tmpfs mount is removed, and files written there won’t be persisted. See [here](https://docs.docker.com/storage/tmpfs/) for the usage and explanation.

One can also mount the host OS's "/tmp"  dir in the container with a bind mount, for exapmle:
```
docker run -v /tmp/tmp_dir:/home/tmp_dir some_image
```
This will cause all data to be written in the host OS's "/tmp/tmp_dir" directory, and the result will depend on how the host manages the "/tmp" directory.

# Our strategy
Occlum is a runtime for Enclave Apps, so it may follow the behavior of the container, it just leaves the "/tmp" directory persisted by default.

# Drawbacks
[drawbacks]: #drawbacks

The files in "/tmp" will not be deleted on a regular interval or after every `occlum run` command.

# Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

Another option is to clean the files after every bootup and on a regular interval, this is more like the traditional Linux. It is more complicated, and the Apps do not expect that "/tmp" is clear before it runs.

liqinggd:
Fixes #570 

Merge https://github.com/occlum/sefs/pull/26 first

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,14 +157,6 @@ Occlum can be configured easily via a configuration file named `Occlum.json`, wh
             "source": "."
         },
         {
-            "target": "/tmp",
-            "type": "sefs",
-            "source": "./run/mount/tmp",
-            "options": {
-                "temporary": true
-            }
-        },
-        {
             "target": "/proc",
             "type": "procfs"
         },
--- a/demos/fish/Occlum.json
+++ b/demos/fish/Occlum.json
@@ -54,14 +54,6 @@
             "source": "."
         },
         {
-            "target": "/tmp",
-            "type": "sefs",
-            "source": "./run/mount/tmp",
-            "options": {
-                "temporary": true
-            }
-        },
-        {
             "target": "/proc",
             "type": "procfs"
         },
--- a/etc/template/Occlum.json
+++ b/etc/template/Occlum.json
@@ -54,14 +54,6 @@
             "source": "."
         },
         {
-            "target": "/tmp",
-            "type": "sefs",
-            "source": "./run/mount/tmp",
-            "options": {
-                "temporary": true
-            }
-        },
-        {
             "target": "/proc",
             "type": "procfs"
         },
--- a/test/Occlum.json
+++ b/test/Occlum.json
@@ -57,14 +57,6 @@
             "source": "."
         },
         {
-            "target": "/tmp",
-            "type": "sefs",
-            "source": "./run/mount/tmp",
-            "options": {
-                "temporary": true
-            }
-        },
-        {
             "target": "/proc",
             "type": "procfs"
         },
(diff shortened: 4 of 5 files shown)


guzongmin (approved):
LGTM

PR #573, opened 2021-08-05
Add similation and debug mode occlum package check and support

qzheng527:

In default only HW mode package is supported. 
But in some cases, such as debug purpose or on no HW SGX env, simulation mode package may be required as well.
Adding a "--debug" flag in "occlum package" to support simulation mode occlum image package.

### DIFF
--- a/tools/occlum
+++ b/tools/occlum
@@ -69,6 +69,8 @@ Usage:
         All runtime dependencies required by the Occlum instance---except Intel SGX driver,
         enable_rdfsbase kernel module, and Intel SGX PSW---are included in the package.
         If package_name is not specified, the directory name of Occlum instance will be used.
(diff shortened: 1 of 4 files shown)


tools/occlum
```diff
@@ -69,6 +69,8 @@ Usage:
         All runtime dependencies required by the Occlum instance---except Intel SGX driver,
         enable_rdfsbase kernel module, and Intel SGX PSW---are included in the package.
         If package_name is not specified, the directory name of Occlum instance will be used.
+        In default only HW mode package is supported. Simulation mode package could be supported
```
guzongmin, line 72: "By default, only HW release mode package is supported"
The below implementation does not check the HW debug mode when the "debug" flag is not set.

guzongmin (approved):
LGTM

PR #574, opened 2021-08-05
Re-organize the structure of procfs and Add support for "/proc/[pid]/stat"

liqinggd:
1. Re-organize the structure of procfs
This commit makes the code is more clear

2. Add support for `/proc/[pid]/stat`


guzongmin (approved):
The /proc/[pid]/stat has a lot of information about the process. In the current implementation, only part of the entry has value. 

But it is good enough now.

PR #575, opened 2021-08-05
Optimize tensorflow_serving workflow

ClawSeven:


tools/docker/ci/Dockerfile.tf_serving
```diff
@@ -0,0 +1,41 @@
+ARG OCCLUM_VERSION
+
+FROM tensorflow/serving:latest-devel as binary_build
```
guzongmin, line 3: Reuse the existing Dockerfile is better than copy it.

ClawSeven: Done

PR #577, opened 2021-08-06
Fix the event_monitor when updating host file events

liqinggd:
When no events happen, the state of host file events will not be reset,
so the event_monitor should always update the state after polling files.


guzongmin (approved):
LGTM

PR #578, opened 2021-08-06
Add creat syscall

liqinggd:


guzongmin (approved):
Align with the description in https://linux.die.net/man/2/creat
"creat() is equivalent to open() with flags equal to O_CREAT|O_WRONLY|O_TRUNC."

PR #583, opened 2021-08-08
Make it accordance for occlum package name and extracted dir name

qzheng527:


guzongmin (approved):
LGTM

PR #584, opened 2021-08-09
Fix exec mode memory leak issue

guzongmin:


src/exec/src/server.rs
```diff
@@ -330,28 +330,36 @@ fn rust_occlum_pal_create_process(
 
     let stdio_raw = Box::new(stdio);
     let mut libos_tid = 0;
-    let create_process_args = Box::new(occlum_pal_create_process_args {
+    let argv = Box::into_raw(cmd_args_array.into_boxed_slice());
+    let env = Box::into_raw(cmd_envs_array.into_boxed_slice());
```
tatetian, line 334: Using `Box` is unnecessary here; so is `into_raw` and `from_raw`. What you want is `Vec::as_ptr`.

```rust
let argv = cmd_args_array.as_ptr();
let env = cmd_envs_array.as_ptr();
```

src/exec/Makefile
```diff
@@ -19,7 +19,7 @@ ifeq ($(OCCLUM_RELEASE_BUILD), 1)
 	@RUSTC_BOOTSTRAP=1 RUSTC_WRAPPER=$(RUSTC_WRAPPER) cargo build --release --target-dir=$(RUST_TARGET_DIR) -Z unstable-options --out-dir=$(RUST_OUT_DIR)
 	@echo "CARGO (release) => exec"
 else
-	@RUSTC_BOOTSTRAP=1 RUSTC_WRAPPER=$(RUSTC_WRAPPER) cargo build --target-dir=$(RUST_TARGET_DIR) -Z unstable-options --out-dir=$(RUST_OUT_DIR)
+	@RUSTC_BOOTSTRAP=1 RUSTC_WRAPPER=$(RUSTC_WRAPPER) RUSTFLAGS=-Zsanitizer=address RUSTDOCFLAGS=-Zsanitizer=address cargo build --target-dir=$(RUST_TARGET_DIR) -Z unstable-options --out-dir=$(RUST_OUT_DIR) -Zbuild-std --target x86_64-unknown-linux-gnu
```
tatetian, line 22: I am not sure that we always want to enable sanitizers (e.g., address sanitizer) even for debug builds. According to my past experience, enabling sanitizers require code instrumentation or other compiler black magics, which means non-trivial runtime cost. Furthermore, the benefits of these sanitizers is minimal since 1) Rust is mostly free from memory safety bugs and 2) the chances of memory leakage problems (except for the trivial ones fixed in this PR) is not high given the fact that the complexity of occlum exec client/server is quite limited. Due to this cost-benefit analysis, **my best judgement is to enable sanitizers only on demand, not by default.**

src/exec/src/server.rs
guzongmin: Thanks for your review comments. Code updated

tatetian (approved):

PR #588, opened 2021-08-11
Fix Cluster Serving demo classpath

qiyuangong:
* Change hard code classpath into path/*, such that changing Flink version will not break this demo.
* Reduce redis memory
* Change Flink version to 1.11.3
* Add no_proxy to avoid network problem in proxy


guzongmin (approved):
LGTM

PR #597, opened 2021-08-12
SGX_SDK environment should not be changed if it is already set

blueskyforeverlove:
If I add SGX_SDK environment to /etc/profile or other environment path then SGX_SDK should not be changed.Maybe the directory which I want to install is not '/opt/intel/'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ install: minimal_sgx_libs
 
 	@echo "Installation is done."
 
-SGX_SDK=/opt/intel/sgxsdk
+SGX_SDK ?= /opt/intel/sgxsdk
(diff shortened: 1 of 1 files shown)


guzongmin:
@blueskyforeverlove I very appreciate you to provide this fix. It looks good. Just one small comments: Can you please change the commit message to start with a verb to align with other commits?
截屏2021-08-13 08 17 57

blueskyforeverlove:
The commit message changed, is that OK?

guzongmin:
@blueskyforeverlove thanks for your contribution

guzongmin (approved):
LGTM

PR #599, opened 2021-08-13
Add SGX v2 self hosted machine for CI

jessehui:


jessehui:
Adding TF serving demo to HW test and also fixing some bugs in #613. Close this.

PR #613, opened 2021-08-25
Add SGX v2 self hosted machine for CI

jessehui:


jessehui:
Verified here: 
Image build: https://github.com/jessehui/occlum/actions/runs/1194051333
Tf-serving-demo: https://github.com/jessehui/occlum/actions/runs/1194050413

Issue #589, opened 2021-08-11
[RFC] Add SGX KSS (Key Separation and Sharing feature) support on Occlum


qzheng527:
- Feature Name: Add SGX KSS (Key Separation and Sharing feature) support on Occlum
- Start Date: 2021-8-11

# Summary
[summary]: #summary

Starting from SGX2, there is a new **Key Separation and Sharing feature** which current Occlum doesn't support . This RFC gives a design proposal about supporting KSS on Occlum.

# Motivation
[motivation]: #motivation

The new **Key Separation and Sharing feature** gives user a chance to fill in some meaningful information to the enclave either in the signing or running stage.

**Signning stage:**
- ISVFAMILYID, 16 bytes
- ISVEXTPRODID, 16 bytes

**Running stage:**
- CONFIG ID, 64 bytes
- CONFIG SVN, 16 bits

All above information could be aquired from the DCAP generated quote in enclave.
Details can be found on the [SGX_Developer_Reference](https://download.01.org/intel-sgx/latest/linux-latest/docs/Intel_SGX_Developer_Reference_Linux_2.14_Open_Source.pdf)
Thus with these user provided information binded with the enclave quote, different attestation and qualification solution can be implemented. 

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

To support the new KSS feature in Occlum, below changes are expected.

1. Add KSS signning stage information in template Occlum.json file.
2. When doing occlum build, the tool gen_internal_conf may check and add the KSS information to SGX signning xml file from Occlum.json file.
3. When doing occlum run, two new added arguments may be usedto pass **CONFIG ID** and **CONFIG SVN** to enclave.

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

- Added KSS part in template Occlum.json.
After `occlum new`, below Occlum.json may be generated.
```
--- a/etc/template/Occlum.json
+++ b/etc/template/Occlum.json
@@ -24,7 +24,12 @@
     "metadata": {
         "product_id": 0,
         "version_number": 0,
-        "debuggable": true
-        "debuggable": true
+        "debuggable": true,
+        "family_id": {
+            "high": "0x0",
+            "low": "0x0"
+        },
+        "ext_prod_id": {
+            "high": "0x0",
+            "low": "0x0"
+        }
     },
```
User can do the changes before `occlum build`.

The occlum tries to parse the string as 64bits hex integer.
For example, enable KSS and set demo IDs as below hex integer.
```
[...]

PR #616, opened 2021-08-30
Add SGX V2 KSS feature support

qzheng527:
Implementaion for RFC (`Add SGX KSS (Key Separation and Sharing feature) support on Occlum`)[https://github.com/occlum/occlum/issues/589](url).

### DIFF
--- a/src/pal/src/pal_enclave.c
+++ b/src/pal/src/pal_enclave.c
@@ -22,6 +22,7 @@
 #include "pal_enclave.h"
 #include "pal_error.h"
 #include "pal_log.h"
+#include "base64.h"
 
 #define MAX_PATH            FILENAME_MAX
 #define TOKEN_FILENAME      "enclave.token"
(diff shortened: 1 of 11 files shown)


src/pal/src/pal_enclave.c
```diff
@@ -51,6 +65,60 @@ static const char *get_enclave_absolute_path(const char *instance_dir) {
...
+        padding = 1;
+    }
+
+    return (len * 3) / 4 - padding;
+}
+
+/**
+ * base64_decode - Base64 decode
+ */
+void base64_decode(const char *b64input, unsigned char *dest, size_t dest_len) {
```
guzongmin, line 87: Please clarify the license of the base64 function.

qzheng527: Will do

tools/gen_internal_conf/src/main.rs
```diff
@@ -286,6 +293,28 @@ fn parse_memory_size(mem_str: &str) -> Result<usize, &str> {
     Ok(mem_val * unit_factor)
 }
 
+fn get_u64_id_high_and_low(id: &OcclumMetaID) -> (u64, u64) {
+    let id_high = u64::from_str_radix(id.high.trim_start_matches("0x"), 16).expect("Hex string ID required");
```
guzongmin, line 297: Ideally, we could use the parser to get the value instead of get it from string.

PR #620, opened 2021-09-03
Modify the statfs on SEFS or UnionFS with ocall

### LINKED ISSUE #586: JAVA file getUsableSpace returns strange values (SEFS)
In developing sofaboot (spring boot), the health check always failed due to the "diskSpace" check.
It returns total disk size as 256k which is much smaller than the threshold 10M.
Tracing the calling process, the issue happened in the flow as below.

**java getUsableSpace->syscall Statfs->...->sefs info.**

From the definition, **the default block size is set to 128, while blocks is set to 128*8=1024**.
(occlum/deps/sefs/rcore-fs-sefs/src/structs.rs)

I don't know why there is such setting which looks to me is not correct.
I tried modifying the block size to 4096, then getUsableSpace may return 100M+ which can satisfy the requirements of the health check. But I don't think it is a good enough solution.

The JAVA code I used to trace the bug.
```
import java.io.File;

public class Main {
   public static void main(String[] args) {
      
      File f = null;
      long v;
      boolean bool = false;
      
      try{
         // create new file
         f = new File(".");
         
         // number of bytes available
         v = f.getUsableSpace();
         
         // true if the file path exists
         bool = f.exists();
         
         // if file exists
         if(bool)
         {
            // prints
            System.out.println("number of bytes available: "+v);
         }
      }catch(Exception e){
         // if any error occurs
         e.printStackTrace();
      }
   }
}
```

BTW, the same java code in Linux host may return the /dev/sda disk size.

liqinggd:
Fixes #586

### DIFF
--- a/src/libos/src/fs/fs_ops/statfs.rs
+++ b/src/libos/src/fs/fs_ops/statfs.rs
@@ -1,11 +1,15 @@
 use super::*;
 use rcore_fs::vfs::FsInfo;
+use std::ffi::CString;
 
 pub fn do_fstatfs(fd: FileDesc) -> Result<Statfs> {
     debug!("fstatfs: fd: {}", fd);
 
     let file_ref = current!().file(fd)?;
-    let statfs = Statfs::from(file_ref.fs()?.info());
+    let statfs = {
+        let fs_info = file_ref.fs()?.info();
+        do_statfs_inner(fs_info)?
+    };
     trace!("fstatfs result: {:?}", statfs);
     Ok(statfs)
 }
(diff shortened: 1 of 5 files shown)


src/libos/src/fs/fs_ops/statfs.rs
```diff
@@ -18,12 +29,23 @@ pub fn do_statfs(path: &str) -> Result<Statfs> {
         let fs = current.fs().read().unwrap();
         fs.lookup_inode(path)?
     };
-    let statfs = Statfs::from(inode.fs().info());
+    let statfs = {
```
guzongmin, line 32: Is it better to combine the do_statfs and do_fstatfs together: one invoke another, or both of them use a common function?

liqinggd: Done

guzongmin (approved):
LGTM

PR #623, opened 2021-09-03
Add SGX2 HW CI and also fix TF-serving demo

jessehui:
Verified here: https://github.com/occlum/occlum/runs/3504529146?check_suite_focus=true


guzongmin (approved):
LGTM

PR #633, opened 2021-09-16
Update sofaboot demo result check command

qzheng527:

For successful sofaboot, every components should be in "UP" status, not "DOWN" status.


guzongmin (approved):
LGTM

PR #638, opened 2021-09-17
Bash

qzheng527:


guzongmin (commented):
We could add two bom files to help user import the bash into their project.

guzongmin (commented):
Dockerfile.aliyunlinux2, Dockerfile.centos7.5 and Dockerfile.centos8.1 are not supported. So we do not need to change them.  Maybe we should remove those files in a separate PR.

qzheng527:
> We could add two bom files to help user import the bash into their project.

Yes, we can do that after this PR got merged.
With the PR merged, the new docker image with prebuilt bash would be pushed to docker hub.
Then the new bom files could be used in the new bash demo.

qzheng527:
> Dockerfile.aliyunlinux2, Dockerfile.centos7.5 and Dockerfile.centos8.1 are not supported. So we do not need to change them. Maybe we should remove those files in a separate PR.

Ok, I will update this PR and create a new PR to remove Dockerfile.centos7.5 and Dockerfile.centos8.1

guzongmin (approved):
LGTM

PR #641, opened 2021-09-18
Enhance autodep with LD_LIBRARY_PATH

StevenJiang1110:
The first commit is about simplifying reading command line options. 
The following two commits is about LD_LIBRARY_PATH. By setting LD_LIBRARY_PATH, we don't need to create softlinks manually. It will find libraries in more directories. 
The last two commits are about inferring dynamic loader for elf file without .interp section and flink demos.


tools/copy_bom/src/util.rs
```diff
@@ -263,11 +310,30 @@ pub fn extract_dependencies_from_output(
...
+                    Some(ref default_lib_dirs) => {
+                        let file_name = (&captures["name"]).to_string();
+                        let lib_dir_in_host = PathBuf::from(&absolute_path)
+                            .parent()
+                            .unwrap()
+                            .to_string_lossy()
+                            .to_string();
+                        // if the shared object is from one of the default dirs, we will copy to the first default dir
+                        // Otherwise it will be copied to the same dir as its dir in host.
+                        if default_lib_dirs.contains(&lib_dir_in_host) {
```
guzongmin, line 326: I think if the first default directory is so special, we should maintain a row for this value.

etc/template/java-11-openjdk-amd64.yaml
```diff
@@ -0,0 +1,20 @@
...
+  # emplicit copy java to enable autodep for java
+  - target: /usr/lib/jvm/java-11-openjdk-amd64/bin
+    copy:
+      - files:
+        - /usr/lib/jvm/java-11-openjdk-amd64/bin/java
+  - target: /usr/lib/jvm/java-11-openjdk-amd64/lib/server
+    copy:
+      - files:
+        - /usr/lib/jvm/java-11-openjdk-amd64/lib/server/libjvm.so
+  - target: /etc
```
guzongmin, line 17: Can you please move this target to line 8, before the copy files?

tools/copy_bom/src/util.rs
StevenJiang1110: fixed. Now, all libraries in these paths will be copied to the loader directory.

etc/template/java-11-openjdk-amd64.yaml
StevenJiang1110: fixed.

guzongmin (approved):
LGTM

PR #646, opened 2021-09-22
Fix the return error code of file operations

liqinggd:
According to https://man7.org/linux/man-pages/man2/write.2.html and https://man7.org/linux/man-pages/man2/read.2.html, If the file is not opened for writing or reading, should return EBADF.


guzongmin (approved):
LGTM

PR #647, opened 2021-09-24
Find dependencies automatically for elf files in copied directories

StevenJiang1110:
Find dependencies for not only copied files, but also files in copied directories.


tools/copy_bom/src/util.rs
```diff
@@ -345,15 +361,19 @@ pub fn extract_dependencies_from_output(
     file_path: &str,
     output: Output,
     default_lib_dirs: Option<Vec<String>>,
+    exit_when_encountering_errors: bool,
```
guzongmin, line 364: I think we should never exit when the share object is not in the default load path, because the so file may in the user folder instead of the common path.

StevenJiang1110: fixed.

guzongmin (approved):
LGTM

PR #668, opened 2021-09-27
Improve userspace VM management

jessehui:
Occlum is a single-address-space library OS. Previously, userspace memory are divided for each process.
And all the memory are allocated when the process is created, which leads to a lot of wasted space and
complicated configuration.

In the current implementation, the whole userspace is managed as a memory pool that consists of chunks. There
are two kinds of chunks:
(1) Single VMA chunk: a chunk with only one VMA. Should be owned by exactly one process.
(2) Multi VMA chunk: a chunk with default chunk size and there could be a lot of VMAs in this chunk. Can be used
by different processes.

This design can help to achieve mainly two goals:
(1) Simplify the configuration: Users don't need to configure the process.default_mmap_size anymore. And multiple processes
running in the same Occlum instance can use dramatically different sizes of memory.
(2) Gain better performance: Two-level management(chunks & VMAs) reduces the time for finding, inserting, deleting, and iterating.


jessehui:
Python musl test is failing because it uses mremap, which will be supported later.

src/Enclave.edl
```diff
@@ -103,6 +103,18 @@ enclave {
...
+         *
+         * Userspace VM manager is a global variable, and it will not be freed until exit.
+         * Thus, freeing the userspace VM needs a ECALL before destroying the enclave.
+         *
+         * @retval On success, return 0. On error, return -errno.
+         *
+         * The possible values of errno are
+         *      EAGAIN - The LibOS is not initialized.
+         */
+        public int occlum_ecall_free_user_space();
```
guzongmin, line 117: Did not get the point why you need to add this ecall? Can you please do the same thing in the enclave destructor function?

jessehui: User space start address is recorded only in trust part. Thus, I think PAL doesn't know the start address of the userspace. ECALL is a way to know this, or I can try `sgx_get_rsrv_mem_info` since there is only one reserved memory area.

PR #710, opened 2021-11-01
Update bash and fish demo with prebuilt busybox

qzheng527:


guzongmin (approved):
LGTM

PR #711, opened 2021-11-01
Make sure bash is built as PIE for differernt OS

qzheng527:

### DIFF
--- a/tools/toolchains/bash/build.sh
+++ b/tools/toolchains/bash/build.sh
@@ -13,7 +13,7 @@ git clone -b ${bash_branch} ${bash_source}
 
 echo "Building bash with musl-gcc (occlum-gcc) ..."
 cd bash
-CC=occlum-gcc CXX=occlum-g++ ./configure --without-bash-malloc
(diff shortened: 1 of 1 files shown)


tools/toolchains/bash/build.sh
```diff
@@ -26,7 +26,7 @@ git clean -dxf
 git reset HEAD --hard
 
 echo "Building bash with gcc ..."
-./configure
+CC="gcc -fPIE -pie" CXX="g++ -fPIE -pie" ./configure
```
guzongmin, line 29: Can you please confirm that "-pie" correct? I remember it is "-fpie".

qzheng527: "-fPIE" is for GCC compile, "-pie" is for link, usually they are used together.
And "-fPIE" is almost the same with "-fpie".
Details please see [`link`](https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html)

guzongmin (approved):
LGTM

Issue #716, opened 2021-11-03
[BUG] Running an application with a large number of arguments will fail


jessehui:
# Describe the bug
When running an application with a large number of arguments in Occlum, the process creation will fail. However, after changing the process' default stack size to a very big value (128MB), there is no difference at all and it still fails.

# To reproduce
The command below is to run a gvisor test in Occlum ignoring all the known failure cases to just run passed cases:

`
[...]

PR #717, opened 2021-11-03
Fix running user application with too many arguments

jessehui:
Fix: #716

### DIFF
--- a/src/libos/src/process/do_spawn/mod.rs
+++ b/src/libos/src/process/do_spawn/mod.rs
@@ -233,7 +233,12 @@ fn new_process_common(
             };
             let user_stack_base = vm.get_stack_base();
             let user_stack_limit = vm.get_stack_limit();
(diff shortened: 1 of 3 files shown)


src/libos/src/process/do_spawn/mod.rs
```diff
@@ -233,7 +233,9 @@ fn new_process_common(
             };
             let user_stack_base = vm.get_stack_base();
             let user_stack_limit = vm.get_stack_limit();
-            let user_rsp = init_stack::do_init(user_stack_base, 4096, &argv, envp, &mut auxvec)?;
+            let init_stack_size = vm.get_stack_range().size() >> 8; // Use 1/256 of user stack range as init size (by default, 4M user stack -> 16K init size)
```
guzongmin, line 236: I think using min(max(vm.get_stack_range().size() >> 8, 4096), vm.get_stack_range().size()) is more reasonable.

jessehui: Done. Thanks for your review.

guzongmin (approved):
LGTM

PR #724, opened 2021-11-08
Create composite action for occlum container run and src build/install

qzheng527:
Abstract the common steps, occlum container creation, source build and install into one composite action.
In this way, much repetitive code could be removed.


guzongmin (approved):
LGTM

PR #761, opened 2021-12-16
Fix the compile error when including dcap_quote.h in pure C program

qzheng527:
Without the fix, the default dcap demo couldn't pass build.


guzongmin (approved):
LGTM

PR #767, opened 2021-12-21
doc(demos/PyTorch): fix wrong message in README

pullp:
There are two errors in PyTorch demo's README.md

- Step 2 and 3 should be executed inside the container, not on host.
- The demo's dir is ` /root/demos/pytorch`, not ` /root/occlum/demos/pytorch`


guzongmin:
@pullp Thanks for your patch.

guzongmin (approved):
LGTM

PR #769, opened 2021-12-24
Fix TCGETS/TCSETS using wrong termios type definition

jessehui:
This implementation is different from what was done in NGO due to different ioctl implementation.


guzongmin (approved):
LGTM

Issue #40, opened 2019-03-21
Collect multithreading requirement from apps


tatetian:
Investigate the use of multithreading in target apps or benchmarks, identify the multithreading requirements, and prioritize the development tasks acccordingly.

PR #776, opened 2022-01-11
BUG: incompatible timespec convertion from rust to c

amazingAmy:
Cannot get correct time when running java/c program in occlum due to incompatible timespec convertion from Rust to C. Actually, the ability of getting correct time is very important for some applications, for example, using JAVA FileLock which repetitively check file last-change time. This bug would make those application rush to error.
The bug is caused by incompatible timespec conversion in Rust ({i64, i32}) and C ({long, long}). When C program invokes a Rust lib to get one variable (e.g., info) including  timespec-type value (info.atime), the result in C  is not same to that generated in Rust.
Remember to check [#40](https://github.com/occlum/sefs/pull/40#issue-1098863814) which also includes one sample code.
In additional, thanks to @liqinggd for his generous help.

### DIFF
--- a/src/libos/src/fs/hostfs.rs
+++ b/src/libos/src/fs/hostfs.rs
@@ -343,15 +343,15 @@ impl IntoFsMetadata for fs::Metadata {
             blocks: self.st_blocks() as usize,
             atime: Timespec {
                 sec: self.st_atime(),
-                nsec: self.st_atime_nsec() as i32,
+                nsec: self.st_atime_nsec(),
             },
             mtime: Timespec {
                 sec: self.st_mtime(),
-                nsec: self.st_mtime_nsec() as i32,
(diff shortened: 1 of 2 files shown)


liqinggd (changes requested):
Thanks for the contribution, Please change the git config with a valid user name and email address.

BTW, there are many codes that use `Timespec`, you should change them all.

guzongmin:
@amazingAmy Thanks for your PR. The CI for this PR failed, I think it may be caused by the SEFS submodule which is not updated at the meantime.

liqinggd:
Please squash the 2 commits together.

guzongmin (commented):
The Enclave_private.pem update should not be part of the fix. Please remove the modification.

liqinggd:
LGTM

guzongmin (approved):
LGTM

PR #779, opened 2022-01-14
[libos] Change ROOT_INODE to ROOT_FS

liqinggd:
The filesystem should be the owner of its root inode, so we should record the fs in occlum

Remember to merge https://github.com/occlum/sefs/pull/42


guzongmin (approved):
LGTM

PR #781, opened 2022-01-17
Enable the grpc ra-tls demo

yuanwu2017:
Signed-off-by: yuanwu <yuan.wu@intel.com>


demos/ra_tls/configs/etc/sgx_default_qcnl.conf
```diff
@@ -0,0 +1,6 @@
+# PCCS server address
+# PCCS_URL=https://localhost:8081/sgx/certification/v3/
+PCCS_URL=https://sgx-dcap-server.cn-shanghai.aliyuncs.com/sgx/certification/v3/
```
qzheng527, line 3: @yuanwu2017 You don't need update this PCCS_URL manually.
In our github action, the composite HW action sets the PCCS_URL as expected.
https://github.com/occlum/occlum/blob/master/.github/workflows/composite_action/hw/action.yml#L43

But for local test, you can add note in the README that the PCCS_URL may be updated according to user's PCCS settings.

qzheng527:
@yuanwu2017 Can you rebase your patch to the latest? Then we can manually trigger HW action CI to see the result?
It takes really long time to do the prepare in my local env. (maybe network is not stable)

demos/ra_tls/configs/etc/sgx_default_qcnl.conf
yuanwu2017: Ok

PR #794, opened 2022-01-21
Fix vfork child process not close opened files

jessehui:
Same as NGO PR: https://github.com/occlum/ngo/pull/197


guzongmin (approved):
LGTM

PR #814, opened 2022-02-11
Add DCAP MAA demo

qzheng527:
Draft， do not merge


demos/remote_attestation/maa/gen_quote/gen_maa_json.c
```diff
@@ -0,0 +1,115 @@
...
+    printf("quote size = %d\n", quote_size);
+
+    p_quote_buffer = (uint8_t*)malloc(quote_size);
+    if (NULL == p_quote_buffer) {
+        printf("Couldn't allocate quote_buffer\n");
+        goto CLEANUP;
+    }
+    memset(p_quote_buffer, 0, quote_size);
+
+    uint8_t enclave_held_data[6] = {0x01, 0x02, 0x03, 0x04, 0x05, 0x06};
```
tychobrailleur, line 71: Is there an example somewhere of how this can be retrieved/computed from an app running with Occlum, for example a JVM-based app?

qzheng527: @tychobrailleur Did you mean how to retrive the "enclave_held_data"? 
The "enclave_held_data" in this example is actaully sgx report data which usually is input by the user.
Once the sgx quote is generated, the report data could be parsed from it, just like the example does in https://github.com/qzheng527/occlum/blob/maa/demos/remote_attestation/maa/gen_quote/gen_maa_json.c#L105.

So for a JVM-based app running in Occlum, similar DCAP operation needs to be added to retrieve any information from the SGX quote. And if the DCAP operation is not feasible to add in the app, Occlum recomends to add it in the Occlum init stage. 
```
For every Occlum built application, it starts [init](https://github.com/occlum/occlum/blob/master/tools/init) process first, then starts the real application in RootFS. The default init process just run RootFS integrity check and then load the RootFS where the real application is located.
```
Thus a customized init needs to be provided when building Occlum image. For how to customize the init , please refer to the demo [`init_ra_flow`](https://github.com/occlum/occlum/tree/master/demos/remote_attestation/init_ra_flow).

tychobrailleur: Yes, that's what I meant.  Makes sense, that `occlum init` info is helpful!  thanks @qzheng527

PR #840, opened 2022-03-10
Fix init_ra HW action timing issue

qzheng527:
1. Fix bug in init_ra HW CI.
2. Delete ra_tls HW CI case since it is almost covered in init_ra demo.
3. Add some brief introduction about init_ra in the doc remote_attestation.


guzongmin (approved):
LGTM

PR #844, opened 2022-03-16
[tools]Support chroot inside copy_bom

StevenJiang1110:
Support executing file operations inside rootfs with `copy_bom`. The example usage is `copy_bom -f XXX.yaml --rootfs YYY`, then file operations will be executed like `chroot YYY cp a b`. If rootfs is set, `copy_bom` will require sudo privilege since `chroot` requires it.


StevenJiang1110:
Under test now. Not ready to merge.

PR #845, opened 2022-03-17
Add support for UTIME

Bonjourz:
1. Add syscall support for UTIME, relevant syscalls:
    a. utime
    b. utimes
    c. futimesat
    d. utimensat

2. Add test case for UTIME, other test cases are enabled in https://github.com/occlum/gvisor/pull/11


src/libos/src/fs/file_ops/utimes.rs
```diff
@@ -0,0 +1,192 @@
...
+        atime.validate()?;
+        let mtime = timespec_t::from(utimbuf.mtime);
+        mtime.validate()?;
+        Ok(utimes_timespec_tuple_t { atime, mtime })
+    }
+
+    pub fn from_timeval_ptr(ptr: *const timeval_t) -> Result<utimes_timespec_tuple_t> {
+        from_user::check_array(ptr, 2)?;
+        let atime_ptr = unsafe { ptr.offset(0) };
+        let atime = unsafe { *atime_ptr };
```
guzongmin, line 47: Should we check the atime_ptr with from_user::check_ptr(atime_ptr)?

src/libos/src/fs/file_ops/utimes.rs
```diff
@@ -0,0 +1,192 @@
...
+        let atime = timespec_t::from(atime);
+        atime.validate()?;
+        let mtime_ptr = unsafe { ptr.offset(1) };
+        let mtime = unsafe { *mtime_ptr };
```
guzongmin, line 51: Should we check the mtime_ptr with from_user::check_ptr(mtime_ptr)?

src/libos/src/fs/file_ops/utimes.rs
```diff
@@ -0,0 +1,192 @@
...
+        let mtime = timespec_t::from(mtime);
+        mtime.validate()?;
+        Ok(utimes_timespec_tuple_t { atime, mtime })
+    }
+
+    pub fn from_timespec_ptr(ptr: *const timespec_t) -> Result<utimes_timespec_tuple_t> {
+        from_user::check_array(ptr, 2)?;
+        let atime_ptr = unsafe { ptr.offset(0) };
+        let atime = unsafe { *atime_ptr };
```
guzongmin, line 60: Should we do more checks before using those unsafe points?

src/libos/src/fs/syscalls.rs
```diff
@@ -700,3 +700,45 @@ pub fn do_flock(fd: FileDesc, operation: i32) -> Result<isize> {
     file_ops::do_flock(fd, flock_ops)?;
     Ok(0)
 }
+
+pub fn do_utime(path: *const i8, times_u: *const utimbuf_t) -> Result<isize> {
+    let utimes_tuple = if !times_u.is_null() {
+        Some(utimes_timespec_tuple_t::from_utimbuf_ptr(times_u)?)
+    } else {
+        None
```
guzongmin, line 708: "If times is NULL, then the access and modification times of the file are set to the current time." 
Should we set the utimes_tuple with current time instead of using Option?

src/libos/src/fs/file_ops/utimes.rs
Bonjourz: The `atime_ptr` is checked in line 45: 
```
from_user::check_array(ptr, 2)?;
```

src/libos/src/fs/file_ops/utimes.rs
Bonjourz: The `mtime_ptr` is checked in line 45:
```
from_user::check_array(ptr, 2)?;
```
Should we refactor the code as:
```
...
let atime_ptr = unsafe { ptr.offset(0) };
from_user::check_ptr(atime_ptr)?
let atime = unsafe { *atime_ptr };
...
let mtime_ptr = unsafe { ptr.offset(1) };
from_user::check_ptr(atime_ptr)?
let mtime = unsafe { *mtime_ptr };
...
```
?

src/libos/src/fs/syscalls.rs
Bonjourz: If the `times_u` is null, the `get_utimes(...)` in `file_ops/utimes.rs` will be called with `times : Option<...> = None`, then it will return:
```
Ok(TimespecTuple {
    atime: Some(now),
    mtime: Some(now),
})
```

src/libos/src/fs/file_ops/utimes.rs
Bonjourz: I think the check is sufficient currently...

liqinggd (changes requested):
We should handle the unsafe structures in `sysycalls.rs` and transform them into safe, the code inside file_ops should use safe structures.

Bonjourz:
> We should handle the unsafe structures in `sysycalls.rs` and transform them into safe, the code inside file_ops should use safe structures.

Done

src/libos/src/fs/syscalls.rs
guzongmin: What I mean is: if we could do the check here, then we do not need to use utimes_timespec_tuple_t instead of Option<utimes_timespec_tuple_t> as the parameter of do_utimes.
BTW: Can we use get_utimes here? If we do this, we may be able to return error earlier.

Bonjourz: I have refactored the code, the diff can be seen in https://github.com/occlum/occlum/pull/846/files. If we think it is better, I will apply such changes to current PR.

Bonjourz: Done.

src/libos/src/fs/file_ops/utimes.rs
```diff
@@ -0,0 +1,154 @@
...
+fn timespec_valid(time: timespec_t) -> bool {
+    if (time.nsec() == UTIME_NOW || time.nsec() == UTIME_OMIT) {
+        true
+    } else {
+        time.sec() >= 0 && time.nsec() >= 0 && time.nsec() < 1_000_000_000
+    }
+}
+
+pub fn do_utimes(
+    dirfd: i32,
```
guzongmin, line 87: Is it better to use FileDesc, FsPath as the parameter of the functions in utimes.rs.
You could add a do_utimes function in syscalls.rs to do the translation, or just move this function to syscalls.rs.

src/libos/src/fs/file_ops/utimes.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+fn do_utimes_fd(fd: FileDesc, atime: Utime, mtime: Utime, flags: i32) -> Result<()> {
+    debug!(
+        "utimes_fd: fd: {:?}, atime: {:?}, mtime: {:?}, flags: {:?}",
+        fd, atime, mtime, flags
+    );
+
+    if flags != 0 {
+        return_errno!(EINVAL, "invalid argument");
```
guzongmin, line 114: The error message could be more clear :"parameter "flags" is invalid"

Bonjourz: Done.

src/libos/src/fs/file_ops/utimes.rs
Bonjourz: I add `do_utimes_wrapper()` in `syscalls.rs` since `do_utimes` is already used.

guzongmin (approved):
LGTM

PR #853, opened 2022-03-28
[libos] Add ENOENT&ENAMETOOLONG checks for path name

lucassong-mh:


src/libos/src/fs/syscalls.rs
```diff
@@ -12,6 +12,8 @@ use super::*;
 use config::ConfigMountFsType;
 use util::mem_util::from_user;
 
+pub const PATH_LEN: usize = 255;
```
guzongmin, line 15: Should we change the PATH_LEN name to MAX_PATH_LEN?
BTW: Can you please add the link about how do you know the MAX_PATH_LEN is 255?

PR #854, opened 2022-03-28
[ci] Add Linux_LTP_test to Demo Test

lucassong-mh:


src/libos/src/fs/syscalls.rs
```diff
@@ -12,6 +12,8 @@ use super::*;
 use config::ConfigMountFsType;
 use util::mem_util::from_user;
 
+pub const MAX_PATH_LEN: usize = 255;
```
lucassong-mh, line 15: Ref in Linux:
https://github.com/torvalds/linux/blob/master/include/linux/ext2_fs.h#L23
https://github.com/torvalds/linux/blob/master/fs/ext4/ext4.h#L2275

guzongmin (approved):
LGTM

Issue #863, opened 2022-04-12
[RFC] Support System V shared memory


Bonjourz:
- Feature Name: Support System V Shared Memory
- Start Day: 2022-3-30

# Summary

Support system V shared Memory in Occlum, related syscalls:
- `shmget()`
- `shmat()`
- `shmdt()`
- `shmctl()`

# Motivation
System V shared memory allows processes to share a region a memory. It is an useful and efficient way for inter-process communication.

# Guide-level explanation
All the processes in Occlum are shared with a single page table, since they are acutually a Linux thread. So for each shared memory segment in Occlum, the **base address** and **permission** is **same** among all processes. (We do not consider thread level permission control currently)

Similar to Linux, related syscalls follow the behavior described in the Linux manual page. We show more details as following:
### **shmget()**
`int shmget(key_t key, size_t size, int shmflg);`

As shown in [man page](https://man7.org/linux/man-pages/man2/shmget.2.html), `shmget()` returns the identifier of the System V shared memory segment associated with the value of the argument key. We call the identifier as `shmid`. The `shmid` is regarded as the handler of the segment in the following system calls.

In Occlum, there are a few `shmflg` we don't need to support:
- `SHM_HUGETLB`, `SHM_HUGE_2MB`, `SHM_HUGE_1GB` and `SHM_NORESERVE`
Such flags is related to the huge page configuration and swap space, which is not supported by Occlum's memory management module.

- (TODO) namespace flag

### **shmat()** and **shmdt()**

`void *shmat(int shmid, const void *shmaddr, int shmflg);`

`int shmdt(const void *shmaddr);`

Details are shown in the [man page](https://man7.org/linux/man-pages/man2/shmat.2.html).

`shmat()` attaches the System V shared memory segment identified by
`shmid` to the address space of the calling process. In Occlum, the base address and its size is fixed when creating a new memory segment. So if the `shmaddr` in `shmat()` is not `NULL` or not equal to the base address of the memory segment, it will return the `EINVAL` errno.

Besides, there are a few `shmflg` for `shmat()` we cannot support in Occlum:
- `SHM_RND` and `SHM_LBA`:
Such flags are used to configure the start address of segment. Since the address of the memory segment is set when creating, Occlum does not support such flags.
- `SHM_EXEC` and `SHM_RDONLY`:
Such flags are used to specify the permission for different processes. In Occlum, all the processes shared one page table, so the permission of a memory segment is same among all processes.
- `SHM_REMAP`
[...]

PR #864, opened 2022-04-12
[libos] Add support for shm

Bonjourz:
1. Add support for System V shared memory;
2. Add some tests for it

More details are shown in #863

### DIFF
--- a/src/libos/src/ipc/shm.rs
+++ b/src/libos/src/ipc/shm.rs
@@ -1,0 +1,509 @@
+use super::*;
+
+use crate::fs::FileMode;
+use crate::process::{do_getegid, do_geteuid, gid_t, uid_t, ThreadRef};
+use crate::time::{do_gettimeofday, time_t};
+use crate::vm::{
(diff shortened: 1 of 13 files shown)


src/libos/src/ipc/shm.rs
```diff
@@ -0,0 +1,509 @@
...
+                shm_segments.remove(&shmid);
+            }
+        } else {
+            return_errno!(EINVAL, "cannot find shm by shmid");
+        }
+        Ok(())
+    }
+
+    fn shmctl_ipcstat(&self, shmid: ShmId, buf: Option<&mut shmids_t>) -> Result<()> {
+        let mut shm_segments = self.shm_segments.write().unwrap();
```
Bonjourz, line 317: should be 
```
let shm_segments = self.shm_segments.read().unwrap();
```

Bonjourz: Done.

Bonjourz:
I have listed what tests have been added for the System V shared memory support.
Search "Test Case" in #863 to get more details.

Issue #865, opened 2022-04-12
The RLIMIT_NOFILE is required for opening files


ClawSeven:
The RLIMIT_NOFILE specifies the maximum number of file descriptor that the process can open. Attempts (open, pipe, dup, etc.) to exceed the limit yield the EMFILE error. Currently, there is no limit implemention for opening files in occlum.

Besides, the failue of the [Linux-ltp getdtablesize01 test](https://github.com/linux-test-project/ltp/blob/91fd818dbff1aa3bcefae3747f83e1dd5c1586dd/testcases/kernel/syscalls/getdtablesize/getdtablesize01.c#L83) indicated that RLIMIT_NOFILE is essential for occlum stability. If the process open files limitlessly, occlum may run out of memory and [core dumped](https://github.com/occlum/occlum/runs/5846569536?check_suite_focus=true) may appear.

qzheng527:
@ClawSeven Occlum does setting the RLIMIT_NOFILE as 1024 in default. Detail please check on https://github.com/occlum/occlum/blob/master/src/libos/src/misc/rlimit.rs#L36.
But currently only "soft limit" was set, but no limit for "hard limit" which is `max` in the structure `rlimit_t`.

ClawSeven:
In current occlum, the RLIMIT_NOFILE has been applied in dup, but not set for open and others syscalls.

ClawSeven:
The recommended [implemention](https://github.com/ClawSeven/occlum/commit/7a9066c0dbe6bdc6349c05fa8e39adacb3842cf0) has been PR in my repo. But due to limitation of opening files, some of gvisor tests may fail, including clock_gettime_test, proc_test and socket_ipv4_udp_unbound_loopback_nogotsan_test. Those tests require large amounts of opening files.

However, if we set relatively large value of RLIMIT_NOFILE in the initial stage, linux-ltp creat05 test may consume a lot of time, so that linux-ltp CI may be canceled due to timeout.

PR #868, opened 2022-04-15
Remove Linux-ltp getdtablesize01 syscalls test

ClawSeven:
The Linux-ltp getdtablesize01 opens "/etc/hosts" file as many, which run out of memory if without limitations. Now I have implemented rlimit_nofile limitations for opening files, but it would cause more failure in gvisor tests. The reasons have been listed in the [issue865](https://github.com/occlum/occlum/issues/865). Therefore, for ensuring CI test integrity as quickly, the Linux-ltp getdtablesize01 test is removed.


guzongmin (approved):
LGTM

PR #884, opened 2022-05-07
Add Spark demo and CI

ClawSeven:
- Add spark demo using spark pi
- Add spark demo hw ci test


.github/workflows/hw_mode_test.yml
```diff
@@ -948,3 +948,63 @@ jobs:
...
+        submodules: true
+
+    - name: Checkout code from fork
+      if: ${{ contains(github.event.pull_request.labels.*.name, 'SGX-hardware-test-required') }}
+      uses: actions/checkout@v2
+      with:
+        ref: refs/pull/${{ github.event.pull_request.number }}/merge
+        submodules: true
+    
+    - name: Create container
```
qzheng527, line 977: It would be better to utilize the HW composite action `"- uses: ./.github/workflows/composite_action/hw"`.
For the special docker run flags "-v $(which docker):/usr/bin/docker -v /var/run/docker.sock:/var/run/docker.sock", I think you can refer to [NGO SIM composite action](https://github.com/occlum/ngo/blob/master/.github/workflows/composite_action/sim/action.yml#L15) to add similar ones for Occlum as well.

demos/spark/prepare_env.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Env variables
+container=intelanalytics/bigdl-ppml-trusted-big-data-ml-scala-occlum
+tag=2.1.0-SNAPSHOT
+SPARK_HOME=/opt/spark
+
+# Install openjdk11
+apt-get update
+apt-get install -y openjdk-11-jre
```
qzheng527, line 10: Is it necessary to install "openjdk-11-jre"?

ClawSeven: Yes, because the java-11-openjdk-amd64.yaml requires openjdk-11-jre. The az team uses openjdk-11 to run spark, so I bring into correspondence with their implementions.

.github/workflows/hw_mode_test.yml
```diff
@@ -948,3 +948,63 @@ jobs:
...
+      run: |
+        docker pull occlum/occlum:latest-ubuntu20.04
+        if [[ "${{ matrix.self_runner[1] }}" == "SGX1-HW" ]]; then
```
jessehui, line 980: Then this "if" looks unnecessary since you only test on SGX 2 machine.

.github/workflows/hw_mode_test.yml
```diff
@@ -948,3 +948,63 @@ jobs:
...
+    - name: Checkout code from fork
+      if: ${{ contains(github.event.pull_request.labels.*.name, 'SGX-hardware-test-required') }}
+      uses: actions/checkout@v2
+      with:
+        ref: refs/pull/${{ github.event.pull_request.number }}/merge
+        submodules: true
+    
+    - name: Create container
+      run: |
+        docker pull occlum/occlum:latest-ubuntu20.04
```
jessehui, line 979: It's better to get Occlum version from the source code instead of using the `latest` tag.

.github/workflows/hw_mode_test.yml
```diff
@@ -948,3 +948,63 @@ jobs:
...
+
+    - name: Make install
+      run: docker exec $spark_test bash -c "source /opt/intel/sgxsdk/environment; cd /root/occlum; OCCLUM_RELEASE_BUILD=1 make install"
+
+    - name: Prepare env
+      run: docker exec $spark_test bash -c "cd /root/occlum/demos/spark && ./prepare_env.sh"
+    
+    - name: Run spark pi
+      run: docker exec $spark_test bash -c "cd /root/occlum/demos/spark && ./run_spark.sh pi"
+
```
jessehui, line 1010: Check the file ending here. Looks like something is wrong.

.github/workflows/hw_mode_test.yml
ClawSeven: Done, nice suggestion!

.github/workflows/hw_mode_test.yml
ClawSeven: Done, I have used "- uses: ./.github/workflows/composite_action/hw"

.github/workflows/hw_mode_test.yml
ClawSeven: Done

ClawSeven: Done

.github/workflows/hw_mode_test.yml
ClawSeven: Done

.github/workflows/demo_test.yml
```diff
@@ -727,4 +727,4 @@ jobs:
 
     - name: Run the LTP demo
       run: docker exec ${{ github.job }} bash -c "cd /root/occlum/demos/linux-ltp/ltp_instance;
-            occlum run /opt/ltp/run-ltp.sh -f syscalls-occlum"
```
qzheng527, line 730: Looks no valid change on this file? Why push it?

.github/workflows/hw_mode_test.yml
```diff
@@ -948,3 +948,40 @@ jobs:
       run: |
         docker exec ${{ env.CONTAINER_NAME }} bash -c "cat /root/occlum/build/test/.fail"
         docker stop ${{ env.CONTAINER_NAME }}
+  
+  Spark_test:
+    if: github.event_name == 'push' || ${{ contains(github.event.pull_request.labels.*.name, 'SGX-hardware-test-required') }}
+    runs-on: ${{ matrix.self_runner }}
+    strategy:
+      matrix:
+        self_runner: [[self-hosted, SGX1-HW], [self-hosted, SGX2-HW]]
```
qzheng527, line 957: If only test on SGX2, you can remove SGX1 here.

.github/workflows/hw_mode_test.yml
```diff
@@ -948,3 +948,40 @@ jobs:
...
+      with:
+        container-name: ${{ github.job }}
+        build-envs: 'OCCLUM_RELEASE_BUILD=1'
+        container-flags: '-v $(which docker):/usr/bin/docker -v /var/run/docker.sock:/var/run/docker.sock'
+    
+    - name: Prepare env
+      run: docker exec ${{ env.CONTAINER_NAME }} bash -c "cd /root/occlum/demos/spark && ./prepare_env.sh"
+    
+    - name: Run spark pi
+      run: docker exec ${{ env.CONTAINER_NAME }} bash -c "cd /root/occlum/demos/spark && ./run_spark.sh pi"
```
qzheng527, line 987: In the end, please add "clean step" just like other jobs.

ClawSeven: Done

PR #897, opened 2022-05-24
Simply build&boot process to combine two internal json files as one

qzheng527:
- Simply build&boot process to combine two internal json files as one.
- Update syscall `(MountRootFS = 363)` to support mount and boot pre-generated UnionFS image.
- Add a demo `runtime_boot` to show how.


etc/template/Occlum.json
```diff
@@ -61,14 +61,6 @@
             "target": "/host",
```
guzongmin, line 61: Suggest removing the /host from the template too. The host FS could be an option for the user. The user needs to add it by himself. The reason is that: The hacker may put some script in the host FS and execute it somehow. Or the hacker may place a soft link in the host FS, which may lead to unexpected behavior.

qzheng527: @guzongmin Even if removing the /host from the tempate, the building generated internal .Occlum_sys.json still has the default hostfs mount information in current solution. Are you suggesting that unless user explictely  setting the /host mount, no default hostfs mount should be enabled in building process?

qzheng527: @guzongmin I fully agree that If "Unless user explicitly set the /host mount, no default hostfs mounted" from security pespective. But perhaps some previous projects may meet problems because they thought that there was a default hostfs mounted. I will discuss with the team.

PR #915, opened 2022-06-20
Stop interrupt thread before destroying the enclave when error

jessehui:


src/exec/src/server.rs
```diff
@@ -22,16 +22,65 @@ use std::sync::{Arc, Condvar, Mutex};
...
+    pub fn set_stopped(&mut self) {
+        *self = Self::Stopped
+    }
+
+    pub fn set_running(&mut self) {
+        *self = Self::Running
+    }
+
+    pub fn is_error(&self) -> bool {
+        match self {
```
volcano0dr, line 51: You can use the matches! macro.

jessehui: Done. Thanks for your review.

volcano0dr (approved):

PR #959, opened 2022-07-20
Fix the issue when path is suffixed by "/"

liqinggd:
Rewrite this PR.

and this commit deps on https://github.com/occlum/occlum/pull/974


volcano0dr (approved):
LGTM

PR #965, opened 2022-07-28
update toolchain golang version to 1.18.4

lambotte-lan:
Signed-off-by: Lan Yun <yun.lan@intel.com>

### DIFF
--- a/.github/workflows/demo_test.yml
+++ b/.github/workflows/demo_test.yml
@@ -56,8 +56,11 @@ jobs:
       run: docker exec ${{ github.job }} bash -c "cd /root/occlum/demos/embedded_mode && SGX_MODE=SIM make;
             SGX_MODE=SIM make test"
 
-    - name: Run Golang sqlite test
-      run: docker exec ${{ github.job }} bash -c "cd /root/occlum/demos/golang/go_sqlite/ && SGX_MODE=SIM ./run_go_sqlite_demo.sh"
+    - name: Run Golang v1.16.3 sqlite test
+      run: docker exec ${{ github.job }} bash -c "cd /root/occlum/tools/toolchains/golang && ./build.sh go1.16.3 && cd /root/occlum/demos/golang/go_sqlite/ && SGX_MODE=SIM ./run_go_sqlite_demo.sh"
+
+    - name: Run Golang v1.18.4 sqlite test
+      run: docker exec ${{ github.job }} bash -c "cd /root/occlum/tools/toolchains/golang && ./build.sh go1.18.4 && cd /root/occlum/demos/golang/go_sqlite/ && SGX_MODE=SIM ./run_go_sqlite_demo.sh"
 
     - name: Go Server set up and run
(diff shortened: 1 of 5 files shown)


guzongmin:
The patch looks good to me. Thanks for your contribution.
Two comments:
1. the commit message could be "Update the Golang version to 1.18.4 and use it as the default one."
2. If we could use both 1.16 and 1.18 in the CI test, it could be better. Please update the "https://github.com/occlum/occlum/blob/master/.github/workflows/demo_test.yml" and run one of the Golang tests with both 1.16 and 1.18?

tools/toolchains/golang/build.sh
```diff
@@ -2,6 +2,19 @@
...
+
+if [[ $# -ge 1 ]];then
+case $1 in
+          --help|help)
+               echo  "Usage:$0 [go1.16.3|go1.18.4]"
+               exit
+               ;;
+          "go1.16.3"|"1.16.3"|"v1.16.3")
+               GO_VERSION="1.16.3"
+               ;;
```
guzongmin, line 15: I think we should add the check about 1.18.4.
"go1.18.4"|"1.18.4"|"v1.18.4")
               GO_VERSION="1.18.4"
               ;;

lambotte-lan: ok, I'll add it

.github/workflows/demo_test.yml
```diff
@@ -56,8 +56,11 @@ jobs:
       run: docker exec ${{ github.job }} bash -c "cd /root/occlum/demos/embedded_mode && SGX_MODE=SIM make;
             SGX_MODE=SIM make test"
 
-    - name: Run Golang sqlite test
-      run: docker exec ${{ github.job }} bash -c "cd /root/occlum/demos/golang/go_sqlite/ && SGX_MODE=SIM ./run_go_sqlite_demo.sh"
+    - name: Run Golang v1.16.3 sqlite test
+      run: docker exec ${{ github.job }} bash -c "cd /root/occlum/tools/toolchains/golang && ./build.sh go1.16.3 && cd /root/occlum/demos/golang/go_sqlite/ && SGX_MODE=SIM ./run_go_sqlite_demo.sh"
+
+    - name: Run Golang v1.18.4 sqlite test
+      run: docker exec ${{ github.job }} bash -c "cd /root/occlum/tools/toolchains/golang && ./build.sh && cd /root/occlum/demos/golang/go_sqlite/ && SGX_MODE=SIM ./run_go_sqlite_demo.sh"
```
guzongmin, line 63: I suggest using "./build.sh go1.18.4" to show test motivation clearly.

tools/toolchains/golang/build.sh
```diff
@@ -17,9 +33,9 @@ cd ${BUILD_DIR}
 # Download Golang
 git clone https://github.com/golang/go .
 # Swtich to Golang 1.13.7
```
guzongmin, line 35: Please update the comments to align with the implementation.

guzongmin (approved):
LGTM
Thanks for your contribution.

PR #974, opened 2022-08-09
Fix the max path limit

liqinggd:
After the updates, should merge https://github.com/occlum/sefs/pull/48 first


volcano0dr (approved):
LGTM

PR #975, opened 2022-08-09
[tools] Fix autodep not work due to non-exist dynamic library

KB5201314:
The current autodep function of copy_bom has a drawback: when one of the elf file's dependencies is missing, none of the dependencies of this elf file will be copied. This requirement exists since sometimes we want to manually copy the .so file to the corresponding directory without pre-installing it on the system directory.

This commit resolved the drawback by replacing `ld.so --list program` with `LD_TRACE_LOADED_OBJECTS=1 ld.so program`.

### DIFF
--- a/tools/copy_bom/src/util.rs
+++ b/tools/copy_bom/src/util.rs
@@ -251,20 +251,20 @@ fn command_output_of_executing_dynamic_loader(
     // return the output of the command to analyze dependencies
     match OCCLUM_LOADERS.ld_library_path_envs.get(dynamic_loader) {
         None => {
-            debug!("{} --list {}", dynamic_loader, file_path);
+            debug!("LD_TRACE_LOADED_OBJECTS=1 {} {}", dynamic_loader, file_path);
             Command::new(dynamic_loader)
-                .arg("--list")
                 .arg(file_path)
+                .env("LD_TRACE_LOADED_OBJECTS", "1")
                 .output()
         }
         Some(ld_library_path) => {
             debug!(
-                "LD_LIBRARY_PATH='{}' {} --list {}",
+                "LD_TRACE_LOADED_OBJECTS=1 LD_LIBRARY_PATH='{}' {} {}",
                 ld_library_path, dynamic_loader, file_path
             );
             Command::new(dynamic_loader)
-                .arg("--list")
                 .arg(file_path)
+                .env("LD_TRACE_LOADED_OBJECTS", "1")
                 .env("LD_LIBRARY_PATH", ld_library_path)
                 .output()
         }
@@ -370,6 +370,9 @@ pub fn extract_dependencies_from_output(
     }
     for line in stdout.lines() {
         let line = line.trim();
+        if line.contains(" => not found") {
+            continue;
+        }
         let captures = DEPENDENCY_REGEX.captures(line);
         if let Some(captures) = captures {
             let raw_path = (&captures["path"]).to_string();


guzongmin:
By design, we want to make sure all the required so files are copied into the right place by this tool. 
What I think is: If some files are missing, the tool needs to report the list of the missing files. By default, the tool would return an error if anything is wrong. Then the user would know something is wrong without debugging the application (it is essential for most users). 
I understand your requirement. My suggestion is: Add one more parameter to this tool to ignore the "so is missing error." But by default, the tool would return an error if any file is missing. Do you agree with it?

KB5201314:
Thank you for your reply.

I just found the description of this situation in [the documentation](https://github.com/occlum/occlum/blob/54bc08d09432380132fed643f9009dc17771c63d/tools/copy_bom/README.md):
> For user-defined elf file, it will report error and abort the program if we can't find the dependent shared objects. 

However it seems that currently copy_bom wouldn't stop or report on `ld.so --list` errors. Users wouldn't see the warning message, unless run it with `OCCLUM_LOG_LEVEL=trace`.


Take the case of missing `librats_tls.so.0`:
```
root@c6d9ac249ed4:/usr/share/rats-tls/samples/occlum_workspace_server# ldd ../rats-tls-server 
        linux-vdso.so.1 (0x00007ffeba3c8000)
        librats_tls.so.0 => not found
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6b2a8b3000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f6b2aab4000)
```
with the following `occlum.yaml`:
```
includes:
  - base.yaml
targets:
  - target: /bin
    copy:
      - files:
        - ../rats-tls-server
```

No warning message, it just failed quietly:
```
root@c6d9ac249ed4:/usr/share/rats-tls/samples/occlum_workspace_server# rm -rf image/
root@c6d9ac249ed4:/usr/share/rats-tls/samples/occlum_workspace_server# copy_bom -f ../occlum.yaml --root image --include-dir /opt/occlum/etc/template
root@c6d9ac249ed4:/usr/share/rats-tls/samples/occlum_workspace_server# echo $?
0
root@c6d9ac249ed4:/usr/share/rats-tls/samples/occlum_workspace_server# tree image/
image/
|-- bin
|   `-- rats-tls-server
|-- dev
|-- etc
|-- host
|-- lib
|-- lib64
|   `-- ld-linux-x86-64.so.2
|-- opt
|-- proc
|-- root
|-- sys
`-- tmp

11 directories, 2 files
```


> My suggestion is: Add one more parameter to this tool to ignore the "so is missing error." But by default, the tool would return an error if any file is missing. Do you agree with it?

I agree with you. But perhaps it would be better to put such options in the bom file rather than as command line arguments? We could extend the definition of the bom file by adding a field to [`struct FileWithOption`](https://github.com/occlum/occlum/blob/f0793b0d305865ba43f4e861ef12de06a1b8149d/tools/copy_bom/src/bom.rs#L66) (e.g. `autodep_allow_fail: Option<bool>`) so that we can control it at the file granularity.

guzongmin:
@KB5201314 BTW: Is the existing "autodep: false" feature meet your requirement?

KB5201314:
> @KB5201314 BTW: Is the existing "autodep: false" feature meet your requirement?

No it doesn't. My program's dependencies are as follows
```
root@08bb98847fb8:~# ldd target/debug/entg
        linux-vdso.so.1 (0x00007fffae147000)
        librats_tls.so.0 => not found
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f2350ded000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2350dca000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2350c7b000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f2350c75000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2350a81000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f2351512000)
```

I still need `autodep: true`, to search and copy most dependencies like `libgcc_s.so.1`, `libpthread.so.0`, `libm.so.6`, `libdl.so.2` and `libc.so.6`. But for the one missing dependency (`librats_tls.so.0`) I would like to copy it to `image` dir manually, because we don't want it to be installed globally on the system.

guzongmin:
You do not need to "installed globally on the system." You only need to copy librats_tls.so.0 to the machine and add the path into the environment variable to help copy_bom to find the shared object file. 

But you do find a bug in the copy_bom. It does not return error information when the dependency is missing.

tools/copy_bom/src/util.rs
```diff
@@ -370,6 +370,9 @@ pub fn extract_dependencies_from_output(
     }
     for line in stdout.lines() {
         let line = line.trim();
+        if line.contains(" => not found") {
+            continue;
```
guzongmin, line 374: If we print the missing shared object file's name and return an error immediately, we fix the "silently fail" bug.

KB5201314:
> You do not need to "installed globally on the system." You only need to copy librats_tls.so.0 to the machine and add the path into the environment variable to help copy_bom to find the shared object file.

Setting the `LD_LIBRARY_PATH` is indeed a good idea, thank you for reminding.

tools/copy_bom/src/util.rs
KB5201314: Yes. Since we don't need `autodep_allow_fail`, we don't have to change the `ld.so --list` now. It will report the missing files in it's stderr. 
```
root@aa5bc5828445:/usr/share/rats-tls/samples/occlum_workspace_server# /lib64/ld-linux-x86-64.so.2 --list ../rats-tls-server
../rats-tls-server: error while loading shared libraries: librats_tls.so.0: cannot open shared object file: No such file or directory
```

All we need to do is change the logic here:
https://github.com/occlum/occlum/blob/af516757ca0bf7499afcdbf459d47e994a385269/tools/copy_bom/src/util.rs#L368-L370

I will create another pr to fix this.

Issue #565, opened 2021-07-28
[RFC] Manage the occlum image with bom file: Design Details


StevenJiang1110:
- Feature Name: occlum bom file
- Start Date: 2021-07-28

# Summary
[summary]: #summary

This RFC is about design details of RFC  #542 . Bom file is used to manage occlum image. In this RFC, we explain why we try to introduce bom file into occlum and the design details of bom files.

# Motivation
[motivation]: #motivation

Currently, what files are included in occlum image is not transparent to users. This will have several drawbacks:
1. The current practice to build an image is not flexible. If we only want to run `hello_world` example compiled with musl, we may not want to copy glibc libraries into occlum image. Since users do not know which files will be copied into the image when they execute `occlum new` or `occlum init`, they can't achieve this goal.
2. It is difficult to handle large programs with very complex dependencies. The current image contains only musl and glibc libraries by default. But for java or python programs, there may be many more dependencies. But occlum currently does not provide a suitable way to include these dependencies.
3. Currently ,to find dependencies for elf files and shared objects relies on human labor. It will be rather dirty work when there are complex dependencies. Bom files can be used to automate management of the dependences.

To deal with all above problems, we wish to introduce a bom file to describe the files we try to include in occlum image. Users can know the source of each file in image from this bom file. We can use bom file to copy files into occlum image. The obvious benefits are
1. The users can have clear idea about what is included in bom image so that they can easily customize their image. 
2. It is convenient to build official images by including all dependencies.

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

1. The bom file format
The bom file is used to describe all file entries included in occlum image. It should meet several requirements:
• It should be convenient for users to read and write.
• It can contain comments.
• It can be easy to extend to support other entries.
2. The bom file content
The bom file should have entries to describe directories and files and can include other bom files. More details can be seen in the bom file template below.
```toml
#Uncomment following line to include other bom files. All entries in included bom files will also be copied to the image. 
#include=[base.bom, musl.bom]

#An example of a file entry. There can be multiple file entries.
#src: the path of the source file. 
#dest: where to put the file. 
#hash: optional and default not set. If hash is set, we will check whether the hash of file is changed when we copy file.
#target_executable: optional and default not set. If target_executable is true, we will find all dependencies of this file and copy these files into image.
[[files]]
src = "/etc/hosts"
dest="image/etc/hosts"

[[files]]
src = "/etc/XXX"
dest="XXXXX"

[...]

PR #981, opened 2022-08-13
[tools] Fix silently fail when autodep fails

KB5201314:
This commit fixes the bug we mentioned here: https://github.com/occlum/occlum/pull/975#discussion_r942031420

### DIFF
--- a/tools/copy_bom/src/error.rs
+++ b/tools/copy_bom/src/error.rs
@@ -6,4 +6,4 @@ pub static CREATE_DIR_ERROR: i32 = -4;
 pub static CREATE_SYMLINK_ERROR: i32 = -5;
 pub static COPY_DIR_ERROR: i32 = -6;
 pub static INCORRECT_HASH_ERROR: i32 = -7;
-pub static AUTODEP_ERROR: i32 = -7;
+pub static AUTODEP_ERROR: i32 = -8;
--- a/tools/copy_bom/src/error.rs
+++ b/tools/copy_bom/src/error.rs
@@ -6,3 +6,4 @@ pub static CREATE_DIR_ERROR: i32 = -4;
 pub static CREATE_SYMLINK_ERROR: i32 = -5;
 pub static COPY_DIR_ERROR: i32 = -6;
 pub static INCORRECT_HASH_ERROR: i32 = -7;
+pub static AUTODEP_ERROR: i32 = -7;
--- a/tools/copy_bom/src/bom.rs
+++ b/tools/copy_bom/src/bom.rs
@@ -402,7 +402,7 @@ impl BomManagement {
         debug!("default loader in autodep: {:?}", default_loader);
         for file_autodep in files_autodep.iter() {
             let mut shared_objects =
-                find_dependent_shared_objects(file_autodep, &default_loader);
+                find_dependent_shared_objects(file_autodep, &default_loader, true);
             for (src, dest) in shared_objects.drain() {
                 let dest_path = dest_in_root(root_dir, &dest);
                 // First, we create dir to store the dependency
@@ -490,7 +490,7 @@ impl BomManagement {
         let default_loader = infer_default_loader(&files_in_copied_dirs);
         let mut all_shared_objects = Vec::new();
         for file_path in files_in_copied_dirs.into_iter() {
-            let shared_objects = find_dependent_shared_objects(&file_path, &default_loader);
+            let shared_objects = find_dependent_shared_objects(&file_path, &default_loader, false);
             all_shared_objects.extend(shared_objects);
         }
         // We should not copy shared libraries already in image directory.
--- a/tools/copy_bom/src/util.rs
+++ b/tools/copy_bom/src/util.rs
@@ -1,6 +1,6 @@
 use crate::error::{
-    COPY_DIR_ERROR, COPY_FILE_ERROR, CREATE_DIR_ERROR, CREATE_SYMLINK_ERROR, FILE_NOT_EXISTS_ERROR,
-    INCORRECT_HASH_ERROR, 
+    AUTODEP_ERROR, COPY_DIR_ERROR, COPY_FILE_ERROR, CREATE_DIR_ERROR, CREATE_SYMLINK_ERROR,
+    FILE_NOT_EXISTS_ERROR, INCORRECT_HASH_ERROR,
 };
 use data_encoding::HEXUPPER;
 use elf::types::{ET_DYN, ET_EXEC, Type};
@@ -203,6 +203,7 @@ pub fn calculate_file_hash(filename: &str) -> String {
 pub fn find_dependent_shared_objects(
     file_path: &str,
     default_loader: &Option<(String, String)>,
+    abort_on_failure: bool,
 ) -> HashSet<(String, String)> {
     let mut shared_objects = HashSet::new();
     // find dependencies for the input file
@@ -214,6 +215,24 @@ pub fn find_dependent_shared_objects(
     let (occlum_elf_loader, inlined_elf_loader) = dynamic_loader.unwrap();
     shared_objects.insert((occlum_elf_loader.clone(), inlined_elf_loader));
     let output = command_output_of_executing_dynamic_loader(&file_path, &occlum_elf_loader);
+    if abort_on_failure {
+        match &output {
+            Err(err) => {
+                error!("cannot autodep for {}: {}", file_path, err);
+                std::process::exit(AUTODEP_ERROR);
+            }
+            Ok(output) if !output.status.success() => {
+                error!(
+                    "cannot autodep for {}. {} stderr: {}",
+                    file_path,
+                    output.status,
+                    String::from_utf8_lossy(&output.stderr).trim()
+                );
+                std::process::exit(AUTODEP_ERROR);
+            }
+            _ => {}
+        }
+    }
     if let Ok(output) = output {
         let default_lib_dirs = OCCLUM_LOADERS
             .default_lib_dirs


volcano0dr:
@KB5201314 Thanks for your contribution.
CI test result shows `Local_attestaion` test failed, can you help to fix this?
Error message: 
`[2022-08-15T00:56:02Z ERROR copy_bom::util] cannot autodep for ../AppResponder/responder. exit status: 127 stderr: Error loading shared library libecdh.so: No such file or directory (needed by ../AppResponder/responder)`

guzongmin:
Maybe the BOM file needs an update. This patch exposes the BOM file issue.

qzheng527:
@KB5201314 @guzongmin @volcano0dr with this new patch, some previous working demos will fail on copy_bom stage,  including all the demos/example which clearly specify the library path on BOM file but not setting correct LD_LIBRARY_PATH, such as [local_attestation](https://github.com/occlum/occlum/blob/master/demos/local_attestation/bom.yaml#L14), and [init_ra_flow](https://github.com/occlum/occlum/blob/master/demos/remote_attestation/init_ra_flow/init_ra_client.yaml#L11).
How about just throwing some clear warning messages instead of error exit?
Just let the users to decide which way is better? Either setting correct LD_LIBRARY_PATH or clearly specify the library path on BOM file.

tools/copy_bom/src/error.rs
```diff
@@ -6,3 +6,4 @@ pub static CREATE_DIR_ERROR: i32 = -4;
 pub static CREATE_SYMLINK_ERROR: i32 = -5;
 pub static COPY_DIR_ERROR: i32 = -6;
 pub static INCORRECT_HASH_ERROR: i32 = -7;
+pub static AUTODEP_ERROR: i32 = -7;
```
qzheng527, line 9: @KB5201314 Please use different error value (-7).

guzongmin:
> @KB5201314 @guzongmin @volcano0dr with this new patch, some previous working demos will fail on copy_bom stage, including all the demos/example which clearly specify the library path on BOM file but not setting correct LD_LIBRARY_PATH, such as [local_attestation](https://github.com/occlum/occlum/blob/master/demos/local_attestation/bom.yaml#L14), and [init_ra_flow](https://github.com/occlum/occlum/blob/master/demos/remote_attestation/init_ra_flow/init_ra_client.yaml#L11). How about just throwing some clear warning messages instead of error exit? Just let the users to decide which way is better? Either setting correct LD_LIBRARY_PATH or clearly specify the library path on BOM file.

Thanks. I get the reason now. So the situation is the same as the previous one @KB5201314 PR wants to address. In the local attestation demo, the "responder" depends on the "libecdh.so". The user just adds libecdh.so into the bom file instead of setting the LD_LIBRARY_PATH.

I think using LD_LIBRARY_PATH is better than manually adding the .so file into the bom file. Since there are only two demos has such an issue, can we simply add the LD_LIBRARY_PATH when using the copy_bom command?

qzheng527:
> > @KB5201314 @guzongmin @volcano0dr with this new patch, some previous working demos will fail on copy_bom stage, including all the demos/example which clearly specify the library path on BOM file but not setting correct LD_LIBRARY_PATH, such as [local_attestation](https://github.com/occlum/occlum/blob/master/demos/local_attestation/bom.yaml#L14), and [init_ra_flow](https://github.com/occlum/occlum/blob/master/demos/remote_attestation/init_ra_flow/init_ra_client.yaml#L11). How about just throwing some clear warning messages instead of error exit? Just let the users to decide which way is better? Either setting correct LD_LIBRARY_PATH or clearly specify the library path on BOM file.
> 
> Thanks. I get the reason now. So the situation is the same as the previous one @KB5201314 PR wants to address. In the local attestation demo, the "responder" depends on the "libecdh.so". The user just adds libecdh.so into the bom file instead of setting the LD_LIBRARY_PATH.
> 
> I think using LD_LIBRARY_PATH is better than manually adding the .so file into the bom file. Since there are only two demos has such an issue, can we simply add the LD_LIBRARY_PATH when using the copy_bom command?

Updating the two demos is simple, but I have two concerns.

1. Perhaps some users already used the "adding the .so file into the bom file" way. 

2. LD_LIBRARY_PATH is soemtimes too powerful. Its prioriy precedes the directories defined in /etc/ld.so.conf, and also /lib, /usr/lib. What if there are same name libraries in LD_LIBRARY_PATH but users actually expects the libraries under /lib ?

guzongmin:
> > > @KB5201314 @guzongmin @volcano0dr with this new patch, some previous working demos will fail on copy_bom stage, including all the demos/example which clearly specify the library path on BOM file but not setting correct LD_LIBRARY_PATH, such as [local_attestation](https://github.com/occlum/occlum/blob/master/demos/local_attestation/bom.yaml#L14), and [init_ra_flow](https://github.com/occlum/occlum/blob/master/demos/remote_attestation/init_ra_flow/init_ra_client.yaml#L11). How about just throwing some clear warning messages instead of error exit? Just let the users to decide which way is better? Either setting correct LD_LIBRARY_PATH or clearly specify the library path on BOM file.
> > 
> > 
> > Thanks. I get the reason now. So the situation is the same as the previous one @KB5201314 PR wants to address. In the local attestation demo, the "responder" depends on the "libecdh.so". The user just adds libecdh.so into the bom file instead of setting the LD_LIBRARY_PATH.
> > I think using LD_LIBRARY_PATH is better than manually adding the .so file into the bom file. Since there are only two demos has such an issue, can we simply add the LD_LIBRARY_PATH when using the copy_bom command?
> 
> Updating the two demos is simple, but I have two concerns.
> 
> 1. Perhaps some users already used the "adding the .so file into the bom file" way.
> 2. LD_LIBRARY_PATH is soemtimes too powerful. Its prioriy precedes the directories defined in /etc/ld.so.conf, and also /lib, /usr/lib. What if there are same name libraries in LD_LIBRARY_PATH but users actually expects the libraries under /lib ?

I think there are some ways to address the issues:
1. Perhaps some users already used the "adding the .so file into the bom file" way.
We can put the additional message "Error loading shared library libpython3.7m.so.1.0: No such file or directory (needed by /root/alpine_python/usr/local/bin/python3.7). \n It seems that some required shared object is missed in the LD_LIBRARY_PATH. Please copy the file into one of the folders in the LD_LIBRARY_PATH or add the shared object folder into the  LD_LIBRARY_PATH."

2. LD_LIBRARY_PATH is soemtimes too powerful. 
I am not suggesting replacing the LD_LIBRARY_PATH with a new one. I suggest using "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<one additional folder>" to add one more folder to help the copy_bom find the missed file.

qzheng527:
> 2\. LD_LIBRARY_PATH is soemtimes too powerful.
>     I am not suggesting replacing the LD_LIBRARY_PATH with a new one. I suggest using "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<one additional folder>" to add one more folder to help the copy_bom find the missed file.

I mean the LD_LIBRARY_PATH precedes the default search path defined in /etc/ld.so.conf or  /lib and /usr/lib which are not defined by LD_LIBRARY_PATH.

guzongmin:
> > 2. LD_LIBRARY_PATH is soemtimes too powerful.
> > I am not suggesting replacing the LD_LIBRARY_PATH with a new one. I suggest using "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:" to add one more folder to help the copy_bom find the missed file.
> 
> I mean the LD_LIBRARY_PATH precedes the default search path defined in /etc/ld.so.conf or /lib and /usr/lib which are not defined by LD_LIBRARY_PATH.



> > 2. LD_LIBRARY_PATH is soemtimes too powerful.
> > I am not suggesting replacing the LD_LIBRARY_PATH with a new one. I suggest using "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:" to add one more folder to help the copy_bom find the missed file.
> 
> I mean the LD_LIBRARY_PATH precedes the default search path defined in /etc/ld.so.conf or /lib and /usr/lib which are not defined by LD_LIBRARY_PATH.

I see. Can we do it like:
LD_LIBRARY_PATH=$LD_LIBRARY_PATH:\<one additional folder\> copy_bom xxx 
then it would not impact other apps

qzheng527:
> > > 2. LD_LIBRARY_PATH is soemtimes too powerful.
> > >    I am not suggesting replacing the LD_LIBRARY_PATH with a new one. I suggest using "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:" to add one more folder to help the copy_bom find the missed file.
> > 
> > 
> > I mean the LD_LIBRARY_PATH precedes the default search path defined in /etc/ld.so.conf or /lib and /usr/lib which are not defined by LD_LIBRARY_PATH.
> 
> > > 2. LD_LIBRARY_PATH is soemtimes too powerful.
> > >    I am not suggesting replacing the LD_LIBRARY_PATH with a new one. I suggest using "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:" to add one more folder to help the copy_bom find the missed file.
> > 
> > 
> > I mean the LD_LIBRARY_PATH precedes the default search path defined in /etc/ld.so.conf or /lib and /usr/lib which are not defined by LD_LIBRARY_PATH.
> 
> I see. Can we do it like: LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<one additional folder> copy_bom xxx then it would not impact other apps

Yes, this way can work for most of the case.
But the LD_LIBRARY_PATH still take effect on the lib/bin defines in the bom file.
For example,  there is a bin file defined in a bom file.
This bin file expects some libs (such as libcrypto.so.1.1) in /lib and also one libexample.so in /home/admin/local/lib/.
But what if there is one libcrypto.so.1.1 in /home/admin/local/lib/ as well?
In this case, LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/admin/local/lib/ copy_bom xxx will include the /home/admin/local/lib/libcrypto.so.1.1 instead of the /lib/libcrypto.so.1.1.
I met such kind of issue when using some libs by extracting them from the docker image rootfs.



the bin expects the libs in /lib and also the libexample.so in

guzongmin:
> > > > 2. LD_LIBRARY_PATH is soemtimes too powerful.
> > > >    I am not suggesting replacing the LD_LIBRARY_PATH with a new one. I suggest using "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:" to add one more folder to help the copy_bom find the missed file.
> > > 
> > > 
> > > I mean the LD_LIBRARY_PATH precedes the default search path defined in /etc/ld.so.conf or /lib and /usr/lib which are not defined by LD_LIBRARY_PATH.
> > 
> > 
> > > > 2. LD_LIBRARY_PATH is soemtimes too powerful.
> > > >    I am not suggesting replacing the LD_LIBRARY_PATH with a new one. I suggest using "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:" to add one more folder to help the copy_bom find the missed file.
> > > 
> > > 
> > > I mean the LD_LIBRARY_PATH precedes the default search path defined in /etc/ld.so.conf or /lib and /usr/lib which are not defined by LD_LIBRARY_PATH.
> > 
> > 
> > I see. Can we do it like: LD_LIBRARY_PATH=$LD_LIBRARY_PATH: copy_bom xxx then it would not impact other apps
> 
> Yes, this way can work for most of the case. But the LD_LIBRARY_PATH still take effect on the lib/bin defines in the bom file. For example, there is a bin file defined in a bom file. This bin file expects some libs (such as libcrypto.so.1.1) in /lib and also one libexample.so in /home/admin/local/lib/. But what if there is one libcrypto.so.1.1 in /home/admin/local/lib/ as well? In this case, LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/admin/local/lib/ copy_bom xxx will include the /home/admin/local/lib/libcrypto.so.1.1 instead of the /lib/libcrypto.so.1.1. I met such kind of issue when using some libs by extracting them from the docker image rootfs.

First, I want to explain why I insist on returning an error. In the current Occlum instance building process, we do not give the user enough support to build a correct instance. For those experienced users, it is not a problem, they could read the log file and find the missing files. But for most users, it is a hard task. 
Then, let us find a solution. Actually, the copy_bom tool has another feature (createlinks) to create a soft link in the image. So the user could use the prefix LD_LIBRARY_PATH to add the additional shared objects into the images and then add create a soft link to help the loader inside Occlum to find the shared object.

KB5201314:
Sorry to reply so late.

> I think using LD_LIBRARY_PATH is better than manually adding the .so file into the bom file. Since there are only two demos has such an issue, can we simply add the LD_LIBRARY_PATH when using the copy_bom command?

Things like `LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<one additional folder> copy_bom` won't work, Because copy_bom will overwrite this env variable for `ld.so`. To make it work, we also need to make a change to this code snippet:
https://github.com/occlum/occlum/blob/5c10af738e378c8703e757832b62639befd0107d/tools/copy_bom/src/util.rs#L260-L272 


> But the LD_LIBRARY_PATH still take effect on the lib/bin defines in the bom file.
For example, there is a bin file defined in a bom file.
This bin file expects some libs (such as libcrypto.so.1.1) in /lib and also one libexample.so in /home/admin/local/lib/.
But what if there is one libcrypto.so.1.1 in /home/admin/local/lib/ as well?
In this case, LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/admin/local/lib/ copy_bom xxx will include the /home/admin/local/lib/libcrypto.so.1.1 instead of the /lib/libcrypto.so.1.1.
I met such kind of issue when using some libs by extracting them from the docker image rootfs.

I think a simple solution to this problem is to split the bom.yaml file into multiple parts and run copy_bom multiple times.

> Then, let us find a solution. Actually, the copy_bom tool has another feature (createlinks) to create a soft link in the image. So the user could use the prefix LD_LIBRARY_PATH to add the additional shared objects into the images and then add create a soft link to help the loader inside Occlum to find the shared object.

Do you mean that the user needs to add a symlink to the additional shared objects file in bom.yaml?

guzongmin:
We do need to discuss the design goal of the copy_bom tool. https://github.com/occlum/occlum/issues/565 is the original design

What I believe are: 
1. The BOM file should not be split. Because it is the best way to help users maintain the content of the Occlum.
2. The tool should report all the missed dependencies. Because it is the best place to report those issues.

guzongmin:
"For example, there is a bin file defined in a bom file.
This bin file expects some libs (such as libcrypto.so.1.1) in /lib and also one libexample.so in /home/admin/local/lib/.
But what if there is one libcrypto.so.1.1 in /home/admin/local/lib/ as well?
In this case, LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/admin/local/lib/ copy_bom xxx will include the /home/admin/local/lib/libcrypto.so.1.1 instead of the /lib/libcrypto.so.1.1."

For this case, my opinion is that: The assumption "what if there is one libcrypto.so.1.1 in /home/admin/local/lib/ as well" is very strange. If the libcrypto.so.1.1 in /home/admin/local/lib/ is not a workable shared object, why would the user put it in the PATH? And we can find another case: the libcrypto.so.1.1 in /lib/ is not a valid one, and the user provided in  /home/admin/local/lib/ is the lib user want to use inside in Occlum. In order to solve such an issue, the copy_bom tool should print detail information about the source path.

qzheng527:
> If the libcrypto.so.1.1 in /home/admin/local/lib/ is not a workable shared object, why would the user put it in the PATH?
For example, the user get most of the libs required but missed several libs whose building may take very long time.
So the user pull a docker image (included the missed libs) and unpack it in some place such as /home/admin/rootfs/.
The missed libs are in the /home/admin/rootfs/lib but there is one unexpected libcrypto.so.1.1 in /home/admin/rootfs/lib as well.

The case is not important, I think the key point here is do we need give below option to the users?
"The copy_bom can work (warning messages if any, let the user do the judgement) by clearly specifying the libs in the bom file but not setting all the library path by LD_LIBRARY_PATH."
 My opinion is it should work otherwise why we call the tool copy_**bom**?

KB5201314:
> > If the libcrypto.so.1.1 in /home/admin/local/lib/ is not a workable shared object, why would the user put it in the PATH?
> > For example, the user get most of the libs required but missed several libs whose building may take very long time.
> > So the user pull a docker image (included the missed libs) and unpack it in some place such as /home/admin/rootfs/.
> > The missed libs are in the /home/admin/rootfs/lib but there is one unexpected libcrypto.so.1.1 in /home/admin/rootfs/lib as well.
> 
> The case is not important, I think the key point here is do we need give below option to the users? "The copy_bom can work (warning messages if any, let the user do the judgement) by clearly specifying the libs in the bom file but not setting all the library path by LD_LIBRARY_PATH." My opinion is it should work otherwise why we call the tool copy_**bom**?

I like the idea of having the bom.yaml file describe everything, including describing a missing lib in bom.yaml. Actually, the test that failed did the same thing (It added the missing libecdh.so to bom.yaml).
https://github.com/occlum/occlum/blob/5c10af738e378c8703e757832b62639befd0107d/demos/local_attestation/bom.yaml#L1-L14

Maybe we should add a field to the bom file to allow autodep to copy the discovered libs, but leave the missing libs to the user.

StevenJiang1110:
Sorry to reply so late. 
It is **very difficult** for `copy_bom` to find all dependencies by setting LD_LIBRARY_PATH. For example, for projects that depend on JDK, we only need to copy a JDK folder at present. But in fact, the libraries that executable files in JDK directories depend on may exist in multiple directories in the JDK folder. If you want to find all dependent libraries through LD_LIBRARY_PATH, you must add all the paths containing these libraries to LD_LIBRARY_PATH, which will be a long list. 

Of course this issue reveals existing problems in copy_bom. When copy_bom cannot find the library, it will generate warning information, but it will only be seen when the log is opened. For the case where the log is not opened, these warnings are directly skipped. This is something that can be improved.

However, before opening the warning, it is necessary to filter the warning information. For the case where some libraries are not found, copy_bom needs to determine whether these libraries are explicitly declared in the bom file, so the libraries in warning message is indeed missed, but this It is not easy to implement, because copy_bom does not record which libraries are copied when copying a folder, and there can be libraries with same name but different content.

tools/copy_bom/src/error.rs
KB5201314: Fixed now :)

StevenJiang1110 (commented):
Thanks for your contribution. I think it makes sense to abort copy_bom when autodep for  **copyfiles** fails (as the PR indicates, ***NOT*** including files in copydirs). However, before aborting copy_bom, we should check whether missing libraries are manually added in bom file. By this way, this PR will not affect current demos.

guzongmin:
> Thanks for your contribution. I think it makes sense to abort copy_bom when autodep for **copyfiles** fails (as the PR indicates, _**NOT**_ including files in copydirs). However, before aborting copy_bom, we should check whether missing libraries are manually added in bom file. By this way, this PR will not affect current demos.

@StevenJiang1110 Thanks for your comments. It seems a good solution. Do you know what is the complexity to add the final check before returning the error?

guzongmin:
> Thanks for your contribution. I think it makes sense to abort copy_bom when autodep for **copyfiles** fails (as the PR indicates, _**NOT**_ including files in copydirs). However, before aborting copy_bom, we should check whether missing libraries are manually added in bom file. By this way, this PR will not affect current demos.

@StevenJiang1110 Thanks for your comments. It seems a good solution. Do you know what is the complexity to add the final check before returning the error?

StevenJiang1110:
> @StevenJiang1110 Thanks for your comments. It seems a good solution. Do you know what is the complexity to add the final check before returning the error?

I think the most complicated case is that user may include these libraries in **copydirs**, but `copy_bom` itself won't check the content of **copydirs**. I think there may be two possible solutions. First is that `copy_bom` itself check all libraries in **copydirs**. Second is to require user to add these libraries in **copyfiles**, then `copy_bom` won't need to check each **copydir**. I'm not sure which choice is more user-friendly.

guzongmin:
> > @StevenJiang1110 Thanks for your comments. It seems a good solution. Do you know what is the complexity to add the final check before returning the error?
> 
> I think the most complicated case is that user may include these libraries in **copydirs**, but `copy_bom` itself won't check the content of **copydirs**. I think there may be two possible solutions. First is that `copy_bom` itself check all libraries in **copydirs**. Second is to require user to add these libraries in **copyfiles**, then `copy_bom` won't need to check each **copydir**. I'm not sure which choice is more user-friendly.

I more like the second one. Users should use copy_bom to build the instance. It may need some time to learn. But it would help the user save a lot of debugging time to find the missing files. I think it is worth learning copy_bom.

qzheng527:
How about we do this way?
Let copy_bom do all based on the bom file, when some libraries are not found, just record them but not quit, continue till complete processing the bom file. Then, check if the recorded missing libraries are in the final image. If yes, copy_bom pass. If not,  return error, and print what libraries are missing.

KB5201314:
> How about we do this way? Let copy_bom do all based on the bom file, when some libraries are not found, just record them but not quit, continue till complete processing the bom file. Then, check if the recorded missing libraries are in the final image. If yes, copy_bom pass. If not, return error, and print what libraries are missing.

How do we check if the missing library is in the oclum image? By checking if the file with the corresponding filename exists in **copyfiles** list?

qzheng527:
> > How about we do this way? Let copy_bom do all based on the bom file, when some libraries are not found, just record them but not quit, continue till complete processing the bom file. Then, check if the recorded missing libraries are in the final image. If yes, copy_bom pass. If not, return error, and print what libraries are missing.
> 
> How do we check if the missing library is in the oclum image? By checking if the file with the corresponding filename exists in **copyfiles** list?

All the generated files (if not quit) are in the final occlum image dir defined by `"--root"` which it is `occlum_instance/image` in default. If there are missing libraries recorded, the final step is to check if they are already in the `occlum_instance/image` (checking the filenames only). If all found, treat the copy_bom successful. If not, error return, print the missing libraries.

guzongmin:
> > > How about we do this way? Let copy_bom do all based on the bom file, when some libraries are not found, just record them but not quit, continue till complete processing the bom file. Then, check if the recorded missing libraries are in the final image. If yes, copy_bom pass. If not, return error, and print what libraries are missing.
> > 
> > 
> > How do we check if the missing library is in the oclum image? By checking if the file with the corresponding filename exists in **copyfiles** list?
> 
> All the generated files (if not quit) are in the final occlum image dir defined by `"--root"` which it is `occlum_instance/image` in default. If there are missing libraries recorded, the final step is to check if they are already in the `occlum_instance/image` (checking the filenames only). If all found, treat the copy_bom successful. If not, error return, print the missing libraries.

I think copy_bom should not depend on the files in the image folder because the original folder maybe not be empty when using the copy_bom tool. The uncertainty would impact the reproducibility. I remember that the copy_bom tool would catch the files copied in a list. So I think we only need to check the missed files in the list before return failure.

guzongmin:
I believe the copy_bom tool is essential because it links the user with Occlum. The BOM file is the interface, and it works well as expected. We need to make the tool better, easy to use, stable, and report precise error messages to help users adjust the BOM file. 

Thanks for @KB5201314 submitting this PR and the developers who care about the tool to discuss improving this tool.

We already discussed a lot in the PR. Maybe someone could write an RFC to discuss how to improve the copy_bom tool before we write the code.

guzongmin:
Fixed in https://github.com/occlum/occlum/pull/1003
Close this PR.
@KB5201314 Thank you very much. I hope you can bring more PRs to help the Occlum project.

Issue #944, opened 2022-07-07
[BUG] Non-deterministic MRENCLAVE value for the same TCB


JasonVranek:
<!--- Provide a general summary of the issue in the Title above -->
# Describe the bug
I'm unsure if this is a bug or if it's an intentional part of the `occlum build` command, but I am unable to generate two signed enclave binaries with the same MRENCLAVE value.

# To reproduce
1. Inside the docker container, follow instructions to run the hello_c demo.
2. Running `occlum build` on the same unmodified source code always yields a different MRENCLAVE value.

# Expected behavior
I expect that if I do not modify the contents of `hello_world.c` (or it's compiled binary), then each enclave instance should produce the same deterministic MRENCLAVE value. 

# Logs
```
root@2feb83e9e875:~# cd demos/hello_c                                                                                                                                                
root@2feb83e9e875:~/demos/hello_c# l                                                                                                                                                 
CMakeLists.txt  Makefile  README.md  hello.yaml  hello_world.c  test.sh*                                                                                                             
root@2feb83e9e875:~/demos/hello_c# make                                                                                                                                              
occlum-gcc hello_world.c -o hello_world                                                                                                                                              
root@2feb83e9e875:~/demos/hello_c# l                                                                                                                                                 
CMakeLists.txt  Makefile  README.md  hello.yaml  hello_world*  hello_world.c  test.sh*                                                                                               
root@2feb83e9e875:~/demos/hello_c# cat test.sh                                                                                                                                       
#!/bin/bash                                                                                                                                                                          
mkdir occlum_workspace                                                                                                                                                               
pushd occlum_workspace                                                                                                                                                               
occlum init                                                                                                                                                                          
rm -rf image                                                                                                                                                                         
copy_bom -f ../hello.yaml --root image --include-dir /opt/occlum/etc/template                                                                                                        
occlum build                                                                                                                                                                         
occlum run /bin/hello_world                                                                                                                                                          
sgx_sign dump -enclave build/lib/libocclum-libos.signed.so -dumpfile metadata.txt                                                                                                    
printf "\n MRENCLAVE: "                                                                                                                                                              
sed -n -e '/enclave_hash.m/,/metadata->enclave_css.body.isv_prod_id/p' ./metadata.txt |head -3|tail -2|xargs|sed 's/0x//g'|sed 's/ //g'                                              
popd                                                                                                                                                                                                                                                                                                                                                                      
rm -r occlum_workspace                                                                                                                                                               

[...]

Issue #992, opened 2022-09-02
[BUG] How to get the static MRENCLAVE in remote attestation demo?


Jiax-cn:
# Description
I added some code in the demo of epid in remote attestation to show the MRENCLAVE in the sgx_report. It worked but everytime I rebuilt this demo, I got a different MRENCLAVE value. I wonder if there is a static MRENCLAVE and why the MRENCLAVE changed.

# To reproduce
1. run in the docker
 `cd demos/remote_attestation/epid`
2. add following code in `demos/remote_attestation/epid/lib/src/ra_quote.cpp > GetEnclaveIasReport` line 195-199
  ```cpp
    uint8_t *mrenclave = quote_args_.quote.as_quote->report_body.mr_enclave.m;
    printf("epid ra mrenclave: ");
    for(int i = 0; i < SGX_HASH_SIZE; i ++)
      printf("%02x ", mrenclave[i]);
    printf("\n");
  ```
3. `./download_and_build.sh`
4. repeat `./run_on_occlum.sh`

# Expected behavior & logs
The output changed.

`epid ra mrenclave: 6e 26 10 9c b4 ab 92 c5 03 89 8c f3 2d 76 02 2c 5d 83 d3 ba 22 11 91 44 99 07 8a 70 c9 c5 2a 4a `

`epid ra mrenclave: c5 63 b2 7f c4 f6 ad 43 d8 04 ad 7f 22 ef a0 74 3a 5f 86 42 59 44 76 e9 cf 1d 3f 92 05 7c 98 34 `

```
root@jiax:~/demos/remote_attestation/epid# ./run_on_occlum.sh 
/root/demos/remote_attestation/epid/occlum_workspace initialized as an Occlum instance
Enclave sign-tool: /opt/occlum/sgxsdk-tools/bin/x64/sgx_sign
Enclave sign-key: /opt/occlum/etc/template/Enclave.pem
SGX mode: HW
rm -rf /root/demos/remote_attestation/epid/occlum_workspace/build
Building new image...
[+] Home dir is /root
[+] Open token file success! 
[+] Token file valid!
[+] Init Enclave Successful 439439578890242!
Generate the SEFS image successfully
Building the initfs...
[+] Home dir is /root
[+] Open token file success! 
[+] Token file valid!
[+] Init Enclave Successful 439508298366978!
Generate the SEFS image successfully
Building libOS...
Signing the enclave...
<EnclaveConfiguration>
    <ProdID>0</ProdID>
[...]

PR #996, opened 2022-09-05
Fix the non-deterministic MRENCLAVE issue when Linux ASLR is enabled

liqinggd:
This PR may fix issues #944 and #992. 

When Linux ASLR is enabled, the padding bytes on the stack are generated randomly, which makes the MAC of the SEFS image differs.

We change the struct `DiskINode` of SEFS to `u64`-aligned to avoid the bytes padding.

Remember to merge this commit https://github.com/occlum/sefs/pull/49 first


volcano0dr (approved):
LGTM

Issue #991, opened 2022-09-01
[RFC] Abort copy_bom when copy_bom fails to find dependencies


StevenJiang1110:
- Feature Name: abort_copy_bom
- Start Date: 2022-09-01

# Summary
[summary]: #summary

Abort `copy_bom` when `copy_bom` fails to find dependencies for **copyfiles** defined in bom files. 

# Motivation
[motivation]: #motivation

`copy_bom` was introduced in #565 to manage files that will be copied into occlum image. `copy_bom` will try to find all dependent libraries for executable files defined in bom files. If `copy_bom` cannot find such libraries, it will generate a warning log message now. However, log is by default closed unless users manually set environmental variable `OCCLUM_LOG_LEVEL`. As a result, even if some libraries are missing, `copy_bom` will exit sliently as indicated in #981. It may misleading users into thinking that all dependencies have been found since users may rely on `copy_bom` to manage dependencies.

# Guide-level explanation
[guide-level-explanation]: #guide-level-explanation

This rfc tries to abort `copy_bom` when dependent libraries for executable files (**only** files defined in **copyfiles**, **not** including files in **copydirs**) are missing. Users can clearly realize that some libraries are missing. `copy_bom` will also output a list of missing libraries so users can try to manage these libraries by themselves, e.g., add the missing libraries to bom file. 

Current occlum users should not realize this feature since their images are successfully built. 

# Reference-level explanation
[reference-level-explanation]: #reference-level-explanation

After discussion in #981, we have reached to several agreements. `copy_bom` can track all missing libraries (the library name) when copying executable files. Before `copy_bom` exits, `copy_bom` will do a final check if all missing libraries are contained in occlum image. 

The final check can be done in several ways. For example, use the `find` command to check each missing library in the occlum image directory.

# Drawbacks
[drawbacks]: #drawbacks

There‘s no clear drawback now.

# Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

- `copy_bom` tries to find missing libraries by itself. This requires users to set correct LD_LIBRARY_PATH. 
- `copy_bom` checks missing libraries in copyfiles in bom file, instead of finding libraries in image directory. See [comment](https://github.com/occlum/occlum/pull/981#issuecomment-1231641031).

# Unresolved questions
[unresolved-questions]: #unresolved-questions

- Is it enough to list library names only for users to manage missing libraries? Since there can be libraries with the same name but with different versions, e.g., musl version or gcc version.

PR #1003, opened 2022-09-09
[tools] Fix copy_bom sliently exit when autodep fails

StevenJiang1110:
This PR is about RFC #991. Similar to #981, but do additional check before copy_bom aborts.

### DIFF
--- a/tools/copy_bom/src/util.rs
+++ b/tools/copy_bom/src/util.rs
@@ -579,3 +579,22 @@ pub fn lazy_check_missing_libraries(image_dir: &str) {
         std::process::exit(MISSING_LIBRARY_ERROR);
     }
 }
+
+/// Print warn message if image directory is not empty.
+/// The image dir should be empty before running copy_bom so copy bom can track all files to copy in bom file.
+/// Users can ignore the warning now if they are sure the image directory contains correct contents.
+pub fn warn_on_nonempty_image_dir(image_dir: &str) {
+    let image_path_buf = PathBuf::from(image_dir);
+    let image_path = image_path_buf.as_path();
+    if image_path.is_dir() {
+        if let Ok(read_dir) = image_path.read_dir() {
+            let dir_entries = read_dir.collect::<Vec<_>>();
+            if dir_entries.len() > 0 {
+                println!(
+                    "WARNING: {} is not an empty directory before running copy_bom.",
+                    image_dir
+                );
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


tools/copy_bom/src/util.rs
```diff
@@ -495,3 +531,51 @@ fn deal_with_output(output: Output, error_number: i32) {
...
+fn add_missing_library(missing_library: String) {
+    MISSING_LIBRARIES
+        .lock()
+        .expect("Acquire lock should not fail")
+        .insert(missing_library);
+}
+
+/// Check whether missing library exists in image directory
+fn check_missing_library(missing_library: &str, image_dir: &str) -> bool {
+    let mut command = Command::new("find");
```
guzongmin, line 545: Can we create a copied_file list to maintain those files already added into the image_dir? Then we could search the missing files in the copied_list.
The difference is: that the copied_list is consistent; it would not change if the image_dir is not clean at the beginning. It could help the user generate the correct and self-consistent bom file.

StevenJiang1110: `copy_bom` can only track files defined in **copyfiles** now, ***NOT*** including files in **copydirs**. The process to copy directory is controlled by rsync only. So, if users pack libraries in a directory, `copy_bom` cannot track such libraries.

guzongmin: > `copy_bom` can only track files defined in **copyfiles** now, _**NOT**_ including files in **copydirs**. The process to copy directory is controlled by rsync only. So, if users pack libraries in a directory, `copy_bom` cannot track such libraries.

Get it, good point. As mitigation, can we print a warning message if the target folder  (i.e. image folder) is not empty at the beginning?

StevenJiang1110: Yes, I will fix it to give a warning message.

tools/copy_bom/src/util.rs
```diff
@@ -495,3 +531,70 @@ fn deal_with_output(output: Output, error_number: i32) {
...
+/// The image dir should be empty before running copy_bom so copy bom can track all files to copy in bom file.
+/// Users can ignore the warning now if they are sure the image directory contains correct contents.
+pub fn warn_on_nonempty_image_dir(image_dir: &str) {
+    let image_path_buf = PathBuf::from(image_dir);
+    let image_path = image_path_buf.as_path();
+    if image_path.is_dir() {
+        if let Ok(read_dir) = image_path.read_dir() {
+            let dir_entries = read_dir.collect::<Vec<_>>();
+            if dir_entries.len() > 0 {
+                warn!(
```
guzongmin, line 593: So only if the user enables the log messages would appear.

StevenJiang1110: Fixed.

guzongmin (approved):
LGTM

qzheng527:
LGTM and verified.

Issue #1009, opened 2022-09-19
[BUG] copy_bom misleading error message


mythi:
# Describe the bug
when running `copy_bom` without `rsync` installed, the error does not give the real reason:
```
[2022-09-19T13:30:07Z ERROR copy_bom::util] copy dir /foo to image /bar failed. No such file or directory (os error 2)
```

# Expected behavior
Error message that says `rsync` cannot be run.

qzheng527:
@mythi Thanks for the info. We will fix this misleading error message.

mythi:
I think this was fixed

PR #1017, opened 2022-09-27
Fix copy_bom error message if rsync is not installed

StevenJiang1110:
Fix issue #1009

### DIFF
--- a/tools/copy_bom/src/util.rs
+++ b/tools/copy_bom/src/util.rs
@@ -1,6 +1,6 @@
 use crate::error::{
     COPY_DIR_ERROR, COPY_FILE_ERROR, CREATE_DIR_ERROR, CREATE_SYMLINK_ERROR, FILE_NOT_EXISTS_ERROR,
-    INCORRECT_HASH_ERROR, MISSING_LIBRARY_ERROR,
+    INCORRECT_HASH_ERROR, MISSING_LIBRARY_ERROR, RSYNC_NOT_FOUND_ERROR,
 };
 use data_encoding::HEXUPPER;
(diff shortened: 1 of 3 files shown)


tools/copy_bom/src/util.rs
```diff
@@ -598,3 +595,17 @@ pub fn warn_on_nonempty_image_dir(image_dir: &str) {
...
     }
 }
+
+/// Check rsync is installed by running command `rsync --version`. 
+/// The exit code should be zere if rsync is installed.
+/// If rsync is not installed, copy_bom will abort.
+pub fn check_rsync()  {
+    let mut command = Command::new("rsync");
+    command.arg("--version");
+    let status = command.status().expect("execute `rsync --version` failed");
```
guzongmin, line 605: Is the rsync info printed out when the user running the copy_bom tool? Should we use "rsync --version > /dev/null"?

StevenJiang1110: Yes, it will print to console. Fixed now.

guzongmin (dismissed):
LGTM

tools/copy_bom/src/util.rs
volcano0dr: If rsync is not installed, copy_bom will panic. Should it output the message "rsync is not installed" and then exit the process with the `RSYNC_NOT_FOUND_ERROR` exitcode instead of having a panic?

StevenJiang1110: Fixed now.

volcano0dr (approved):
LGTM

PR #1022, opened 2022-10-10
Make vfork stop parent child threads

jessehui:
When vfork is called and the current process has other running child threads, for Linux, the other threads remain running. For Occlum, this behavior is different. All the other threads will be frozen until the vfork returns or execve is called in the child process. The reason is that since Occlum doesn't support fork, many applications will use vfork to replace fork. For multi-threaded applications, if vfork doesn't stop other child threads, the application will be more likely to fail because the child process directly uses the VM and the file table of the parent process.


volcano0dr (approved):
LGTM

PR #1035, opened 2022-10-25
Add get_dcap_quote

hzjane:
When Applying EHSM remote attestation in occlum. Need to get dcap quote and send it to EHSM for attestation.
To write quote to a binary file，and then running PPML jar will read it and send it to attest.

### DIFF
--- a/demos/remote_attestation/dcap/c_app/dcap_c_test.c
+++ b/demos/remote_attestation/dcap/c_app/dcap_c_test.c
@@ -48,7 +48,7 @@ void dump_quote_info(sgx_quote3_t *p_quote)
     printf("\t0x%04x\n", p_rep_body->config_svn);
 }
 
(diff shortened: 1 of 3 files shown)


demos/remote_attestation/dcap/c_app/dcap_c_test.c
```diff
@@ -69,7 +69,11 @@ void main() {
     memset(p_quote_buffer, 0, quote_size);
 
     sgx_report_data_t report_data = { 0 };
-    char *data = "ioctl DCAP report data example";
+    char *data = "ppml";
+    if (args[1] != NULL && args[1] != "") {
+    	data = args[1];
```
guzongmin, line 74: Using unchecked data as report data is a security bug. Because a hacker could use this interface to generate any reports.  It is not a good demo for users.

PR #1042, opened 2022-11-02
[sefs] Configure larger cache size to improve pfs performance

lucassong-mh:
This PR configure **cache size** of underlying SGX-PFS of SEFS to **1MiB** (default 192KiB).

Larger cache size can acclerate file I/O of SEFS.
(20% ~ 70% reads/writes improvemenet according to FIO test)


volcano0dr (approved):
LGTM

PR #1058, opened 2022-11-30
[demos] Move all the benchmark demos together

qzheng527:
Make Occlum/NGO have the same benchmark demos location to ease AliOS CI development.


volcano0dr (approved):
LGTM

PR #1065, opened 2022-12-07
[toolchain] Make occlum-go support gcc/glibc option

qzheng527:
In default, occlum-go uses musl-gcc. Users can configure occlum-go to use gcc like below.
" CC=gcc occlum-go build "


volcano0dr (approved):
LGTM

PR #1066, opened 2022-12-09
[demos] Add MySQL demo

lucassong-mh:
This PR adds MySQL demo to Occlum's demo list.
Run step:
1. Download mysql-8.0.31 and build dependencies
2. Build from src (Need to apply patch to disable `HAVE_TIMES`)
3. Run mysql server and test mysql client


volcano0dr (approved):
LGTM

PR #1074, opened 2022-12-14
Improve clock_gettime, stream socket and add MySQL demo

lucassong-mh:
This PR includes:

1. Workaround `CLOCK_PROCESS_CPUTIME_ID` and `CLOCK_THREAD_CPUTIME_ID` in `clock_gettime()`. The syscall with either these two flags would end up calling `vdso_ocall_clock_gettime()`.
2. Improve `ioctl()` in stream socket: Refine state transfer.
3. Add MySQL to demo list.


jessehui (approved):
LGTM

PR #1098, opened 2022-12-29
Fix brk not initialize memory

jessehui:


src/libos/src/vm/process_vm.rs
```diff
@@ -440,25 +439,40 @@ impl ProcessVM {
...
+            // Get page-aligned brk address.
+            let new_brk = align_up(brk, PAGE_SIZE);
+            // Get page-aligned old brk address.
+            let old_brk = align_up(*brk_guard, PAGE_SIZE);
+
+            // Reset the memory when brk shrinks.
+            if new_brk < old_brk {
+                let shrink_brk_range =
+                    VMRange::new(new_brk, old_brk).expect("shrink brk range must be valid");
+                USER_SPACE_VM_MANAGER.reset_memory(shrink_brk_range)?;
```
volcano0dr, line 462: Should brk return `Ok(*brk_guard)` when `reset_memory` returns Error?

jessehui: I think not. If there is an error during `reset_memory`, there must be a fatal error. Just returning `Ok(*brk_guard)` might cover the error.

volcano0dr (approved):
LGTM

PR #1111, opened 2023-01-12
Improve benchmark ci and hw ci

lucassong-mh:
1. Enable benchmark ci on schedule(Wed,Sat at 10PM UTC (6AM CST)) only on branch 1.0.0-preview
2. Improve hw ci: a. run on schedule(Tue,Fri at 6PM UTC (2AM CST)) b. lower stress test times under given limit(360 min)


volcano0dr (approved):
LGTM

PR #1112, opened 2023-01-12
[async-socket] Support close and shutdown Netlink socket

ClawSeven:


volcano0dr (commented):
If the Domain is Netlink, there is no need to call host_shutdown.
https://github.com/occlum/occlum/blob/1.0.0-preview/src/libos/crates/async-socket/src/datagram/sender.rs#L218

volcano0dr (approved):
LGTM

PR #1115, opened 2023-01-12
[ci] Improve schedule in benchmark ci and hw ci

lucassong-mh:
1. Disable benchmark test on push
2. Improve schedule of benchmark ci and hw ci


volcano0dr (approved):
LGTM

PR #1116, opened 2023-01-13
[ci] Do not run container with network host for hw ci

qzheng527:
To avoid network port conflict between different runners.


volcano0dr (approved):
LGTM

PR #1119, opened 2023-01-13
[libos] Fix vulnerable nonce in DCAP verifier

KB5201314:
This commit fixed a security issue in the dcap verifier. The issue was caused by the `[in, out]` attributes of pointer parameter `qve_report_info` in the ocall function `occlum_ocall_verify_dcap_quote()`. This led to the vulnerability where the protected `qve_report_info.nonce` field in libos could be arbitrarily rewritten by attacker outside libos.

Here is a simple proof of the attack.

For illustration purposes, we add two statements to quote_verifier.rs, which will print the contents of qve_report_info.nonce before and after the call to oclum_ocall_verify_dcap_quote(). Then we added a statement in src/pal/src/ocalls/attestation.c to act as an attacker on the ocall side who will modify the nonce value.
```patch
diff --git a/src/libos/src/util/sgx/dcap/quote_verifier.rs b/src/libos/src/util/sgx/dcap/quote_verifier.rs
index c37fb4ae..56adf529 100644
--- a/src/libos/src/util/sgx/dcap/quote_verifier.rs
+++ b/src/libos/src/util/sgx/dcap/quote_verifier.rs
@@ -48,6 +48,7 @@ impl QuoteVerifier {
             if sgx_status != sgx_status_t::SGX_SUCCESS {
                 return_errno!(EAGAIN, "failed to get random number from sgx");
             }
+            error!("qve_report_info.nonce.rand: {:X?}", qve_report_info.nonce.rand);
         }
 
         qve_report_info.app_enclave_target_info = get_self_target()?;
@@ -66,6 +67,7 @@ impl QuoteVerifier {
                 supplemental_data.as_mut_ptr(),
             );
             assert_eq!(sgx_status_t::SGX_SUCCESS, sgx_status);
+            error!("qve_report_info.nonce.rand: {:X?}", qve_report_info.nonce.rand);
         }
 
         match qe3_ret {
diff --git a/src/pal/src/ocalls/attestation.c b/src/pal/src/ocalls/attestation.c
index 13c456bd..71f184f4 100644
--- a/src/pal/src/ocalls/attestation.c
+++ b/src/pal/src/ocalls/attestation.c
@@ -138,6 +138,7 @@ quote3_error_t occlum_ocall_verify_dcap_quote(
     uint8_t *supplemental_data
 ) {
 #ifndef OCCLUM_DISABLE_DCAP
+    *(uint32_t*)(&qve_report_info->nonce.rand[0]) = 0xdeadbeef;
     return sgx_qv_verify_quote(
                quote_buf, quote_size,
                (sgx_ql_qve_collateral_t *)quote_collateral,
```

We can verify this attack by running the demo under `demos/remote_attestation/dcap`:
```sh
cd demos/remote_attestation/dcap
OCCLUM_LOG_LEVEL=error ./run_dcap_quote_on_occlum.sh
```
output:

It can be seen that the nonce values within libos are modified by attacker and not detected by `sgx_tvl_verify_qve_report_and_identity()`.

After applying this commit, although the attacker can modify the nonce value used to generate QvE report, it can be detected by `sgx_tvl_verify_qve_report_and_identity()`.

output:

### DIFF
--- a/src/libos/src/util/sgx/dcap/quote_verifier.rs
+++ b/src/libos/src/util/sgx/dcap/quote_verifier.rs
@@ -39,6 +39,7 @@ impl QuoteVerifier {
         let mut collateral_expiration_status = 1;
         let mut supplemental_data = vec![0; self.supplemental_data_size as usize];
         let mut qve_report_info = sgx_ql_qe_report_info_t::default();
+        let mut nonce;
 
         unsafe {
             let sgx_status = sgx_read_rand(
@@ -48,6 +49,7 @@ impl QuoteVerifier {
             if sgx_status != sgx_status_t::SGX_SUCCESS {
                 return_errno!(EAGAIN, "failed to get random number from sgx");
             }
+            nonce = qve_report_info.nonce;
         }
 
         qve_report_info.app_enclave_target_info = get_self_target()?;
@@ -66,6 +68,9 @@ impl QuoteVerifier {
                 supplemental_data.as_mut_ptr(),
             );
             assert_eq!(sgx_status_t::SGX_SUCCESS, sgx_status);
+            // We have to re-write qve_report_info.nonce with the value we backed up earlier,
+            // since qve_report_info.nonce can be overwrite by attacker from ocall side.
+            qve_report_info.nonce = nonce;
         }
 
         match qe3_ret {


volcano0dr (approved):
Thanks for reporting this issue, LGTM

PR #1121, opened 2023-01-16
Fix vulnerable nonce in DCAP verifier

qzheng527:
This commit fixed a security issue in the dcap verifier. The issue was caused by the '[in, out]' attributes of pointer parameter qve_report_info in the ocall function occlum_ocall_verify_dcap_quote(). This led to the vulnerability where the protected qve_report_info.nonce field in libos could be arbitrarily rewritten by attacker outside libos.

Signed-off-by: Kun Lai <me@imlk.top>


volcano0dr (approved):
LGTM

PR #1128, opened 2023-01-30
Fix mremap lock range when merging connecting chunks

jessehui:
Test with the command `TESTS=mmap make test times=2000` and pass.


volcano0dr (approved):
LGTM

PR #1155, opened 2023-02-16
Pg support

qzheng527:
Some changes required to support PG15 running in Occlum


src/libos/src/fs/inode_file.rs
```diff
@@ -113,6 +113,46 @@ impl File for INodeFile {
         Ok(*offset as i64)
     }
 
+    fn preadv(&self, bufs: &mut [&mut [u8]], offset: usize) -> Result<usize> {
+        if !self.access_mode.readable() {
+            return_errno!(EBADF, "File not readable");
+        }
+        self.offset.lock().unwrap();
```
liqinggd, line 120: The lock here seems to be useless because the preadv does not change the seek offset of the opened file.

src/libos/src/fs/inode_file.rs
```diff
@@ -113,6 +113,46 @@ impl File for INodeFile {
...
+            }
+        }
+        Ok(total_len)
+    }
+
+    fn pwritev(&self, bufs: &[&[u8]], offset: usize) -> Result<usize> {
+        if !self.access_mode.writable() {
+            return_errno!(EBADF, "File not writable");
+        }
+        self.offset.lock().unwrap();
```
liqinggd, line 140: The lock here seems to be useless, because the pwritev does not change the seek offset of the opened file.

src/libos/src/fs/syscalls.rs
```diff
@@ -207,6 +208,34 @@ pub fn do_pread(fd: FileDesc, buf: *mut u8, size: usize, offset: off_t) -> Resul
     Ok(len as isize)
 }
 
+pub fn do_preadv(fd: FileDesc, iov: *mut iovec_t, count: i32, offset: off_t) -> Result<isize> {
```
liqinggd, line 211: Can the `do_preadv` and `do_readv` be combined to reduce the repetitive code?

liqinggd (dismissed):
LGTM

volcano0dr (approved):
LGTM

PR #1158, opened 2023-02-20
[glibc] No need build libcrypt from glibc

qzheng527:
The libcrypt.so built from glibc doesn't have "XCRYPT_2.0" section.
Using crypt implementation coming from libxcrypt which is installed already in Ubuntu 20.04.


volcano0dr (approved):
LGTM

PR #1160, opened 2023-02-20
[sefs] Make cache size configurable in Occlum.yaml

lucassong-mh:
Similar modification in https://github.com/occlum/occlum/pull/1142
This PR
1. add `sefs_cache_size` to `ConfigMount.options`
2. rename `total_size` to `async_sfs_total_size` to avoid ambiguity


volcano0dr (approved):
LGTM

PR #1181, opened 2023-03-02
[libos] Fix the align_up & align_down assertions in prelude

liqinggd:
The `is_power_of_two()` ensures the alignment is not zero.


volcano0dr (approved):
LGTM

PR #1191, opened 2023-03-07
Add PaddlePaddle demo

henrysun007:


qzheng527:
@henrysun007 Can you add some description of this demo.py in the README?
Looks it is a run check for paddle.
It would be better to add one real case example, such as MNIST inference/training.

qzheng527:
LGTM, thanks for your contribution.

PR #1210, opened 2023-03-17
Fix potential server test failure

jessehui:
Due to scheduling, the client could attempt to connect to the server before the server starts to accept. Make the client retry multiple times before returning with error.


volcano0dr (approved):
LGTM

Issue #1196, opened 2023-03-10
[BUG] app is killed by SIGSEGV shortly after executing the cpuid instruction


KB5201314:
# Describe the bug
I'm working on running [librats](https://github.com/inclavare-containers/librats) on Occlum. And found that my app was always getting killed by a `SIGSEGV` signal.  Through debugging I found that the app was killed is in the `cpuid()` function of librats.

Here is a sample code that reproduces the bug.
As a comparison, I also picked up the `native_cpuid()` function from Occlum's code.
cpuid.c:
```c
#include <sys/stat.h>
#include <sys/sysmacros.h>
#include <fcntl.h>
#include <unistd.h>
#include <stdbool.h>

/* code from occlum start */
// https://github.com/occlum/occlum/blob/e30a8a63e20f335595e3babb9896bb3f41c0acf8/test/cpuid/main.c#L13-L29

typedef struct t_cpuid {
    unsigned int eax;
    unsigned int ebx;
    unsigned int ecx;
    unsigned int edx;
} t_cpuid_t;

static inline void native_cpuid(int leaf, int subleaf, t_cpuid_t *p) {
    memset(p, 0, sizeof(*p));
    /* ecx is often an input as well as an output. */
    asm volatile("cpuid"
                 : "=a" (p->eax),
                 "=b" (p->ebx),
                 "=c" (p->ecx),
                 "=d" (p->edx)
                 : "a" (leaf), "c" (subleaf));
}
/* code from occlum end */


/* code from librats start */
// https://github.com/inclavare-containers/librats/blob/c0654baad9464e0fba2646bdf8abad18e4a8254a/core/cpu.c#L34-L37

static inline void cpuid(int *eax, int *ebx, int *ecx, int *edx)
{
    asm volatile("cpuid"
             : "=a"(*eax), "=b"(*ebx), "=c"(*ecx), "=d"(*edx)
             : "0"(*eax), "1"(*ebx), "2"(*ecx), "3"(*edx)
             : "memory");
}
/* code from librats end */


int main(int argc, char const *argv[])
{
    t_cpuid_t cpu;
    printf("native_cpuid() start\n");
    native_cpuid(0x12 /* SGX_LEAF */, 0, &cpu);
    printf("native_cpuid() end\n");

    int a[4];
    a[0] = 0x12 /* SGX_LEAF */;
    a[2] = 0;
    printf("cpuid() start\n");
    cpuid(&a[0], &a[1], &a[2], &a[3]);
[...]

PR #1197, opened 2023-03-10
[libos] fix user stack content being overwritten

imlk0:
The current implementation of `__occlum_sysret` overwrites the contents of the user stack, which may causes crashes in leaf functions which depend on "red zone".

This commit moves the check for `pku_enabled` ahead of `popfq` to avoid the problem.

However, this problem still exists in the `update_pkru_in_sysret` function (since several registers inevitably need to be backed up before calling wrpkru). This commit provides a workaround by skipping the red zone (128 bytes accroding to x86_64 System V ABI) at the top of the stack.

fix #1196

PR #1212, opened 2023-03-21
[libos] Retain red zone area and FLAGS unmodified across syscall

Bonjourz:
Same as https://github.com/occlum/occlum/pull/1197, fix https://github.com/occlum/occlum/issues/1196.

The pervious implementation of `occlum_sysret` misuses the red zone, which breaches the System V ABI.
At the prologue and epilogue of syscall handling functions, we need to perform the following operations:
1. Check whether to change the value of PRKU register;
2. If the condition in 1. is true, change the value of PKRU register.

These operations may overwrites several registers, such as FLAGS, %rax, %rdx and %rcx. We use the stack area below the red zone to save and restore the register above.


volcano0dr (approved):
LGTM

PR #1213, opened 2023-03-22
[jindisk] Support threaded logging

lucassong-mh:
This PR adds a new feature: threaded logging to JinDisk.

Threaded-logging enables JinDisk to log data in seperate blocks rather big segment. It triggers when disk space is running out. It automatically disables when background GC freed some segments. Foreground GC is no longer needed thanks to threaded logging.


tatetian (approved):

PR #1216, opened 2023-03-23
Rewrite sendfile with async closure

liqinggd:


liqinggd:
The async closure can not modify the mutable variables outside of it.

PR #1231, opened 2023-04-04
[async-socket] Limit the number of listen backlog for async accept request

jessehui:
This should fix the slow shutdown/close for the socket.


volcano0dr (approved):
LGTM

PR #1236, opened 2023-04-11
Retain red zone area and FLAGS unmodified across syscall

Bonjourz:
1. The previous implementation of __occlum_syret usse the red zone to save the register temporarily, which may overwrite the red zone area. This change avoids using the red zone when handling syscall.

2. Save and restore the FLAGS register before and after the operations on PKRU to keep the FLAGS unmodified.


volcano0dr (approved):
LGTM

PR #1244, opened 2023-04-17
Use the default poll implementation of INode

liqinggd:
Please merge https://github.com/occlum/sefs/pull/56 first

### DIFF
--- a/src/libos/src/fs/dev_fs/dev_fd.rs
+++ b/src/libos/src/fs/dev_fs/dev_fd.rs
@@ -15,14 +15,6 @@ impl INode for DevFd {
         Err(vfs::FsError::PermError)
     }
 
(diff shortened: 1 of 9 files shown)


src/libos/src/fs/dev_fs/dev_random.rs
```diff
@@ -14,14 +14,6 @@ impl INode for DevRandom {
         Err(FsError::PermError)
     }
 
-    fn poll(&self) -> vfs::Result<vfs::PollStatus> {
```
volcano0dr, line 17: For dev/random, the returned `PollStatus::write`should be false.

liqinggd: Ok, the `poll` of inode in devfs may be special.

volcano0dr (approved):
LGTM

PR #1247, opened 2023-04-20
Fix wait4 failure of child process created with vfork

jessehui:


src/libos/src/process/do_wait4.rs
```diff
@@ -54,7 +58,8 @@ pub fn do_wait4(child_filter: &ProcessFilter, options: WaitOptions) -> Result<(p
         .collect::<Vec<&ProcessRef>>();
 
     if unwaited_children.len() == 0 {
-        return_errno!(ECHILD, "Cannot find any unwaited children");
+        // Check again for vfork-ed child process which doesn't have a real structure of a process
+        return wait4_exit_child_created_with_vfork(process.pid(), child_filter);
```
volcano0dr, line 62: If there are child processes and all child processes are running, the vfork-ed child process of the terminated should also be checked.

src/libos/src/process/do_wait4.rs
```diff
@@ -30,6 +31,9 @@ pub fn idle_reap_zombie_children() -> Result<()> {
         }
     }
 
+    // Also check for vforked child process
+    reap_zombie_child_created_with_vfork();
```
volcano0dr, line 35: I don't think this function should be called here, this function should be used to clean up memory when libos is shutdown.

src/libos/src/process/do_wait4.rs
jessehui: Fixed. Thanks for your review.

src/libos/src/process/do_wait4.rs
jessehui: Yes. This is wrong. I re-implement this function and call it in `exit_process`. Please take a new look. Thank you for your review.

src/libos/src/process/do_vfork.rs
```diff
@@ -201,3 +228,62 @@ pub fn handle_force_stop() {
         }
     }
 }
+
+// Wait4 unwaited child which are created with vfork and directly exit without calling execve.
+pub fn wait4_exit_child_created_with_vfork(
+    parent_pid: pid_t,
+    child_filter: &ProcessFilter,
+) -> Result<(pid_t, i32)> {
```
volcano0dr, line 236: I think it makes more sense for the return type of `wait4_exit_child_created_with_vfork` to be `Option<(pid_t, i32)>`,  the error of no child process is handled in `do_wait4`.

src/libos/src/process/do_vfork.rs
```diff
@@ -201,3 +228,62 @@ pub fn handle_force_stop() {
...
+            }
+            return Ok((*child.pid(), child.status().as_u32() as i32));
+        }
+    }
+
+    return_errno!(ECHILD, "no exit child found");
+}
+
+// Reap all unwaited child which are created with vfork and directly exit without calling execve.
+pub fn reap_zombie_child_created_with_vfork(parent_pid: pid_t) -> Vec<pid_t> {
```
volcano0dr, line 258: reap_zombie_child_created_with_vfork(parent_pid: pid_t) -> Option<Vec<pid_t>>

src/libos/src/process/do_vfork.rs
jessehui: Done. Thanks for your review.

src/libos/src/process/do_vfork.rs
jessehui: Done. Thanks for your review.

volcano0dr (approved):
LGTM

PR #1268, opened 2023-05-12
Occlum init-ra initfs

qzheng527:


qzheng527:
Draft， do not merge

tools/init_grpc_ratls/src/main.rs
```diff
@@ -0,0 +1,235 @@
...
+            path: keys.path.clone(),
+            val_buf: buffer.clone(),
+        };
+
+        keys_info.push(key_info);
+    }
+    Ok(keys_info)
+}
+
+fn main() -> Result<(), Box<dyn Error>> {
```
volcano0dr, line 121: fn main() -> std::io::Result<()>

Issue #1281, opened 2023-05-26
Track unsupported/incomplete features of AsyncSFS


lucassong-mh:
AsyncSFS has some unsupported/incomplete features compared to SEFS (discovered by CI):

**Unsupported:**
U1. Modification of inode's metadata by different operations (e.g., chmod, umask)
U2. Create file which size is larger than 4GB

**Incomplete:**
I1. [fallocate()](https://man7.org/linux/man-pages/man2/fallocate.2.html) support, like `FALLOC_FL_PUNCH_HOLE`, `FALLOC_FL_KEEP_SIZE`

**Failed test cases:**

| Test                             | case                                 | cause                     |
| -------------------------------- | ------------------------------------ | ------------------------- |
| make test: mmap                  | test_file_backed_mremap              | I1                        |
|                                  | test_file_backed_mremap_mem_may_move | I1                        |
| make test: file                  | test_fallocate_punch_hole            | I1 (FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE) |
| make test: truncate              | test_truncate_then_read              | I1                        |
| make test: chmod                 | test_chmod                           | U1 (stat.st_mode)         |
| make test: chown                 | test_chown                           | U1 (stat.st_uid&st_gid)   |
| make test: access                | test_access                          | U1 (fstat.st_mode)        |
| make test: umask                 | test_create_file_with_default_umask  | U1 (fstat.st_mode)        |
| make test: utimes                | test_utime                           | U1 (stat.st_atim&st_mtim) |
|                                  |                                      |                           |
| gvisor syscall test: access_test | AccessTest.UsrReadWrite              | U1                        |
| gvisor syscall test: mkdir_test  | MkdirTest.HonorsUmask                | U1                        |
|                                  | MkdirTest.HonorsUmask2               | U1                        |
| gvisor syscall test: utimes_test | UtimesTest.OnFile                    | U1                        |
|                                  | UtimesTest.OnDir                     | U1                        |
|                                  | FutimesatTest.OnAbsPath              | U1                        |
|                                  | FutimesatTest.OnRelPath              | U1                        |
|                                  | UtimensatTest.OnAbsPath              | U1                        |
|                                  | UtimensatTest.OnRelPath              | U1                        |
|                                  | UtimeTest.ZeroAtimeandMtime          | U1                        |
|                                  | Utimensat.NullPath                   | U1                        |

@liqinggd

PR #1282, opened 2023-05-26
[libos] Replace SEFS to AsyncSFS+JinDisk as Occlum's default storage backend

lucassong-mh:
This PR includes:

- AsyncSFS+PageCache+JinDisk are Occlum's default storage backend now. The rootfs(unionfs) is initialized as RO layer: SEFS and RW layer: AsyncSFS+JinDisk. The initfs remains same configuration (RO-SEFS and RW-SEFS) in initialize stage.
- The default `resource_limits.kernel_space_heap_size` is increased since PageCache(256MB by default) and JinDisk(8MB per GB data + 64MB fixedly) would consume more.
- The legacy configuration(SEFS) is preserved since AsyncSFS has some unsupported/incomplete features compared to SEFS, tracked in https://github.com/occlum/occlum/issues/1281
- `occlum mount` is temporarily unusable until jindisk-cli is delivered to replace sefs-cli. `occlum mount` could also be replaced by building a [busybox](https://github.com/mirror/busybox) inside occlum and do `occlum run /bin/busybox [cmd]` to check dirs and file contents.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -110,7 +110,7 @@ jobs:
       if: ${{ always() }}
       run: docker stop ${{ env.CONTAINER_NAME }}
 
-  SEFS_FIO_Test:
+  FIO_Test:
     timeout-minutes: 60
     runs-on: ${{ matrix.self_runner }}
     strategy:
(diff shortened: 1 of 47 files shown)


volcano0dr (approved):
LGTM

PR #1300, opened 2023-06-05
[poll] Fix panic for non-opened fds

jessehui:
Fix a bug recently introduced by the support of negative poll_fd.


volcano0dr (approved):
LGTM

PR #1305, opened 2023-06-08
Fix building TF serving image failure

jessehui:
Due to limited resources of Github Actions' hosted machine, we changed to build this image on our self-hosted machine.


volcano0dr (approved):
LGTM

PR #1289, opened 2023-05-30
[jindisk] Protect RIT with symmetric key cipher

cqs21:
This PR contains:
- Add symmetric key cipher (AES128 CTR mode) api
- Add encryption(symmetric) for DiskArray blocks

Only RIT implementation uses the DiskArray at the moment.

PR #1307, opened 2023-06-12
[jindisk] Improve block cryption, reclaim and discard

lucassong-mh:
This PR includes:
1. Add symm cryption API (AES-128-CTR) and apply to `DiskArray` @cqs21 https://github.com/occlum/occlum/pull/1289
2. Refactor cryption API (refer to [rust-openssl](https://docs.rs/openssl/0.10.54/openssl/symm/index.html))
3. Improve block reclamation part in compaction
4. Rename `trim()` to `discard()` for more accuracy in block device layer. Add an implementation, leave performance impact `TODO`.

### DIFF
--- a/src/libos/crates/jindisk/src/util/cryption.rs
+++ b/src/libos/crates/jindisk/src/util/cryption.rs
@@ -25,28 +25,32 @@ pub type Iv = [u8; AUTH_ENC_IV_SIZE];
 
 /// Encryption/Decryption.
 pub trait Cryption {
-    /// Encrypt a block.
-    fn encrypt_block(plaintext: &[u8], key: &Key) -> CipherBlock;
+    /// Encrypt a block. (AEAD)
+    fn encrypt_block_aead(plaintext: &[u8], key: &Key) -> CipherBlock;
 
-    /// Decrypt a block.
-    fn decrypt_block(ciphertext: &[u8], key: &Key, meta: &CipherMeta) -> Result<[u8; BLOCK_SIZE]>;
+    /// Decrypt a block. (AEAD)
+    fn decrypt_block_aead(
+        ciphertext: &[u8],
+        key: &Key,
+        meta: &CipherMeta,
+    ) -> Result<[u8; BLOCK_SIZE]>;
 
-    /// Encrypt content of any length.
-    fn encrypt_arbitrary(plaintext: &[u8], ciphertext: &mut [u8], key: &Key) -> CipherMeta;
+    /// Encrypt content of any length. (AEAD)
+    fn encrypt_arbitrary_aead(plaintext: &[u8], ciphertext: &mut [u8], key: &Key) -> CipherMeta;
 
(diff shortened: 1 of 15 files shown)


src/libos/crates/jindisk/src/util/disk_array.rs
```diff
@@ -79,9 +84,9 @@ impl<T: Serialize> DiskArray<T> {
     }
 
     pub async fn persist(&self, _root_key: &Key) -> Result<()> {
-        // TODO: Add encryption(symmetric) for blocks
         for (hba, block) in self.table.iter() {
-            self.disk.write(*hba, block.as_slice()).await?;
+            let ciphertext = DefaultCryptor::symm_encrypt_block(block.as_slice(), &self.key)?;
```
tatetian: Is there security weakness in using the same key for every block? I thought the crypto API should take an argument for user to decide the IV. This way, different block can be encrypted with different IV, although the key is the same. Should our crypto API allow providing IVs?

src/libos/crates/jindisk/src/util/cryption.rs
```diff
@@ -25,28 +25,32 @@ pub type Iv = [u8; AUTH_ENC_IV_SIZE];
...
 
-    /// Encrypt a block using symmetric key cipher(AES128 CTR mode)
-    fn symm_encrypt_block(data: &[u8], key: &Key) -> Result<[u8; BLOCK_SIZE]>;
+    /// Encrypt a block. (Non-AEAD)
+    fn encrypt_block(data: &[u8], key: &Key) -> Result<[u8; BLOCK_SIZE]>;
 
-    /// Decrypt a block using symmetric key cipher(AES128 CTR mode)
-    fn symm_decrypt_block(data: &[u8], key: &Key) -> Result<[u8; BLOCK_SIZE]>;
+    /// Decrypt a block. (Non-AEAD)
+    fn decrypt_block(data: &[u8], key: &Key) -> Result<[u8; BLOCK_SIZE]>;
```
tatetian, line 53: The current crypto API encourages copying by returning `[u8; BLOCK_SIZE]`. Usually, the user of such API can prepare the target enc/dec buffer before calling the API. So by accepting `&mut [u8]` as input, the API can save the user from one extra copy of a block of memory.

tatetian (commented):
Is there any performance degradation introduced by this PR?

The PR looks mostly good to me. I raised some minor issues.

src/libos/crates/jindisk/src/util/cryption.rs
lucassong-mh: The start point to design this interface is to ensure the invariant of block-granularity cryption. There is a non-copy version called `encrypt_arbitrary(plaintext, ciphertext)` can serve the purpose.

And no obvious performance degradation according to some microbenchmarks by this PR.

tatetian (approved):

PR #1356, opened 2023-07-26
Optimize unionfs to reduce memory usage

liqinggd:


liqinggd:
This commit can't ensure that the inode id is constant, for example,
if open /A/B/C, the inode id is 5, then close, then reopen /A/B/C, the inode id may not be 5. This behavior is not correct.

PR #1358, opened 2023-07-26
[hw ci] Reconfigure mysql for benchmark on AsyncSFS+JinDisk

lucassong-mh:
Test [here](https://github.com/lucassong-mh/occlum/actions/runs/5666004494/job/15351861146) and [here](https://github.com/lucassong-mh/occlum/actions/runs/5666587066)


volcano0dr (approved):
LGTM

Issue #1392, opened 2023-09-15
[BUG] Rust app in occlum


fixcik:
Hello, Occlum team! We encountered two bugs while working with Rust in Occlum.

1) Mmap
When attempting to read a single file concurrently in multiple threads, we catch unexpected behavior.
> Source code: https://github.com/fixcik/occlum-bug/blob/master/src/main.rs
> Build: https://github.com/fixcik/occlum-bug/blob/master/Dockerfile
> Config: https://github.com/fixcik/occlum-bug/blob/master/Occlum.json

2) Non-null pointer in is_inside_user_space check
An uninitialized vec is pointing to 0x1.
See issue: https://github.com/rust-lang/rust/issues/39625
And when we try to write an empty vec, the error "the whole buffer is not in the user space" occurs.
This fix solves our problem: https://github.com/aggregion/occlum/commit/19586bcfb02580fc40f2d8d82882e9e18f3037f8

fixcik:
Build log:

```
#10 [ 6/11] RUN occlum-cargo build -r && cp target/x86_64-unknown-linux-musl/release/occlum-bug . && rm -rf target
#10 0.606     Updating crates.io index
#10 1.520  Downloading crates ...
#10 2.084   Downloaded memmap v0.7.0
#10 2.269   Downloaded libc v0.2.148
#10 2.381    Compiling libc v0.2.148
#10 3.991    Compiling memmap v0.7.0
#10 4.422    Compiling occlum-bug v0.1.0 (/build)
#10 4.985     Finished release [optimized] target(s) in 4.60s
#10 DONE 5.2s

#11 [ 7/11] RUN occlum new instance
#11 0.486 /build/instance initialized as an Occlum instance
#11 DONE 0.7s

#12 [ 8/11] WORKDIR /build/instance
#12 DONE 0.0s

#13 [ 9/11] COPY Occlum.json .
#13 DONE 0.0s

#14 [10/11] COPY bom.yaml .
#14 DONE 0.0s

#15 [11/11] RUN rm -rf image   && copy_bom -f ./bom.yaml --root image --include-dir /opt/occlum/etc/template   && occlum build;
#15 0.357 Enclave sign-tool: /opt/occlum/sgxsdk-tools/bin/x64/sgx_sign
#15 0.357 Enclave sign-key: /opt/occlum/etc/template/Enclave.pem
#15 0.357 SGX mode: HW
#15 0.380 rm -rf /build/instance/build
#15 0.458 Building the initfs...
[...]

PR #1394, opened 2023-09-19
[vm] Avoid early munmap of shm

lucassong-mh:
This PR fixes an early munmap bug in posix shm, inspired by https://github.com/occlum/occlum/issues/1392

TODO: Add the multi-thread rust demo that could testify the above bugfix


volcano0dr (approved):
LGTM

PR #1409, opened 2023-10-09
Enable copy_bom to find dependences directories which end with '/'

StevenJiang1110:
The original way to get filename is incorrect for directories which end with '/'.

For path like `/dev/pts/`, if we use the original way,
```rust
let pathbuf = PathBuf::from("/dev/pts/");
let filename = pathbuf.filename().unwrap(); 
```
`filename` will be `pts`. However, an empty filename is indeed what we want.

### DIFF
--- a/tools/copy_bom/README.md
+++ b/tools/copy_bom/README.md
@@ -47,7 +47,7 @@ The second part in the line indicates where to find shared libraries. All paths 
 
 # known limitations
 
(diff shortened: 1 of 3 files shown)


volcano0dr:
LGTM,  Please update the readme and example to explain the new usage.

volcano0dr (approved):
LGTM

PR #1410, opened 2023-10-10
Improve network performance by reducing ocalls

ClawSeven:
This PR improve 50% tcp performance by reducing ocalls.

image

### DIFF
--- a/src/libos/src/net/socket/host/send.rs
+++ b/src/libos/src/net/socket/host/send.rs
@@ -1,5 +1,17 @@
 use super::*;
 
+#[derive(Debug)]
+pub struct Sender {
+    alloc: UntrustedSliceAlloc,
+}
+
+impl Sender {
+    pub fn new() -> Result<Self> {
+        let alloc = UntrustedSliceAlloc::new(SEND_BUF_SIZE)?;
+        Ok(Self { alloc })
+    }
+}
+
 impl HostSocket {
     pub fn send(&self, buf: &[u8], flags: SendFlags) -> Result<usize> {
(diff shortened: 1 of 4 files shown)


ClawSeven:
In the previous implementation, Occlum used mm_alloc ocall to allocate an untrusted buffer before performing send/recv ocall. After the completion of the send/recv ocall, Occlum used mm_dealloc ocall to deallocate the untrusted buffer. These implementations resulted in at least three ocalls in one user send/recv syscall, leading to significant transmission overhead.

However, in the current PR implementation, Occlum allocates the untrusted buffer for each socket construction. The lifecycle of the untrusted buffer is bound to the socket's lifecycle, allowing Occlum to reuse these untrusted buffers for each subsequent send/recv syscall within the same socket.

src/libos/src/net/socket/host/send.rs
```diff
@@ -24,16 +36,31 @@ impl HostSocket {
...
             for buf in data {
-                bufs.push(u_allocator.new_slice(buf)?);
+                let u_slice = u_allocator.new_slice(buf)?;
+                bufs.push(u_slice);
             }
             bufs
         };
 
-        self.do_sendmsg_untrusted_data(&u_data, flags, name, control)
+        let bytes = self.do_sendmsg_untrusted_data(&u_data, flags, name, control);
```
volcano0dr, line 61: Please modify `bytes` to `retval`
```
let retval = self.do_sendmsg_untrusted_data(...);
u_allocator.reset();
retval
```

volcano0dr (approved):
LGTM

PR #1418, opened 2023-10-18
Implement untrusted IO buffer for each thread

ClawSeven:


src/libos/src/net/socket/host/mod.rs
```diff
@@ -18,16 +15,13 @@ mod recv;
 mod send;
 mod socket_file;
 
-pub const SEND_BUF_SIZE: usize = 128 * 1024;
-pub const RECV_BUF_SIZE: usize = 128 * 1024;
+pub const SOCK_BUF_SIZE: usize = 128 * 1024;
```
volcano0dr, line 18: I think `SOCK_BUF_SIZE` should be renamed to `IO_BUF_SIZE` and defined in thread.

volcano0dr (approved):
LGTM

PR #1421, opened 2023-10-25
Fix concurrent mprotect causing conflict chunk and vma for Single VMA chunk

jessehui:


src/libos/src/vm/vm_manager.rs
```diff
@@ -313,8 +315,25 @@ impl VMManager {
...
+
+                // There are rare cases that mutliple threads do mprotect or munmap for the same single-vma chunk
+                // but for different ranges and the cloned chunk is outdated when acquiring the InternalVMManger lock here.
+                // In this case, we search for the chunk again.
+                let chunk = if chunk.range() != vma.lock().unwrap().range() {
+                    let current = current!();
+                    let process_mem_chunks = current.vm().mem_chunks().read().unwrap();
+                    let chunk = process_mem_chunks
+                        .iter()
+                        .find(|&chunk| chunk.range().intersect(&protect_range).is_some());
```
volcano0dr, line 329: I'm not sure if this should be: `find(|&chunk| chunk.range().is_superset_of(&protect_range)) `

jessehui: This is more accurate. Fixed. Thanks for your review.

volcano0dr (approved):
LGTM

PR #1451, opened 2023-12-15
Added an Initial version of a NGINX demo

dzobbe:
Hi all,

I started to work on the setup of a NGINX demo in Occlum. Unfortunately, it still does not work completely. I addressed many startup errors and now the servers starts. However, when a client request is made, the clients connect to the server but it hangs. 

I would like to have some help :)

Cheers.


qzheng527:
@dzobbe By my undestanding, the nginx uses **fork** which is not supported by Occlum.
That was also why we didn't add nginx demo.
Is "fork" not necessary in nginx now? Or you can modify the "fork“ with "vfork" or "posix_spawn" which Occlum supports?

dzobbe:
Yes, in fact I modified the fork with vfork to make it compile 
https://github.com/occlum/occlum/pull/1451/files#diff-7141555e4dbf090d6677dbc5a83bf4a54497c169fad03dea9de890a0fae428c7R17

jessehui:
@dzobbe Thanks for your contribution. It looks like the `rt_sigsuspend` is not supported and it looks like an important syscall for Nginx.

dzobbe:
Oh, didn't notice that. So it seems there is no way to make it work? :(

jessehui:
Not really. For `rt_sigsuspend`, it doesn't seem to be an impossible task. We will have a discussion about that. May we know the scenario of your use case or the reason why you want to support Nginx?

BUPT-wxq:
I want to migrate a nginx service to occlum, which is used as communication in occlum. Can ngxin work properly now ?

PR #1455, opened 2023-12-20
[config] Add "feature" field to Occlum.json

lucassong-mh:
This PR includes:

1. A new configurable field "feature" is added in Occlum.json. Two features are now supported: "pkru" and "enable_posix_shm".
2. Minor improvement in shared chunk expansion in shm: Add current process to the shared set when there is a `NeedExpand` for two chunks.
3. Enable all features on hw ci for EDMM.

### DIFF
--- a/src/libos/src/vm/shm_manager.rs
+++ b/src/libos/src/vm/shm_manager.rs
@@ -103,6 +103,7 @@ impl ShmManager {
                                 .unwrap()
                                 .writeback_file()
                                 .unwrap(),
+                            options.perms(),
                         ),
                     ) {
                         return Ok(MmapSharedResult::NeedExpand(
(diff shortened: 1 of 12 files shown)


src/libos/src/vm/shm_manager.rs
```diff
@@ -192,23 +193,16 @@ impl ShmManager {
...
-            // Apply higher perms to the whole new range
-            let new_perms = new_vma.perms();
-            let old_perms = old_vma.perms();
-            if new_perms != old_perms {
-                let perms = new_perms | old_perms;
-                new_vma.set_perms(perms);
-                new_vma.modify_permissions_for_committed_pages(perms);
-            }
+            // Inherit access from the old shared chunk
+            new_vma.inherits_access_from(&old_vma);
```
volcano0dr, line 205:
```
fn merge_two_single_vma_chunks(&mut self, lhs: &ChunkRef, rhs: &ChunkRef) -> ChunkRef {
    let mut new_vma = {
        let lhs_vma = lhs.get_vma_for_single_vma_chunk();
        let rhs_vma = rhs.get_vma_for_single_vma_chunk();
        debug_assert_eq!(lhs_vma.end(), rhs_vma.start());

        let mut new_vma = lhs_vma.clone();
        new_vma.set_end(rhs.end());
        new_vma
    };

    self.update_single_vma_chunk(&current!(), lhs, new_vma)
}
```
This reduces calls to reset access.

lucassong-mh: Make sense and done

volcano0dr (approved):
LGTM

PR #1571, opened 2024-06-24
Add Filebench demo and enable it on CI to test Ext2 

lucassong-mh:
1. Add Filebench tool
2. Run a micro benchmark to test Ext2 on CI


volcano0dr (approved):
LGTM

PR #1573, opened 2024-06-25
Check validity of cache and disk size in mount configs during 'occlum build'

lucassong-mh:


tools/gen_internal_conf/src/main.rs
```diff
@@ -300,6 +300,52 @@ fn main() {
         );
         assert!(kernel_heap_max_size >= kernel_heap_init_size);
 
+        // Check validity of cache and disk size in mount options
```
volcano0dr, line 303: This check code can be moved to line 478.

tools/gen_internal_conf/src/main.rs
```diff
@@ -300,6 +300,52 @@ fn main() {
...
+                    println!(
+                        "Invalid cache_size \"{}\". The cache_size must be 1. larger than the minimum size \"{}\", \
+                        2. aligned with cache page size \"{}\", \
+                        3. smaller than the kernel_heap_max_size \"{}\".",
+                        cache_size, MIN_CACHE_SIZE, CACHE_PAGE_SIZE, kernel_heap_max_size
+                    );
+                    return;
+                }
+            }
+            if let Some(disk_size_str) = mount.options.disk_size.as_ref() {
```
volcano0dr, line 330: Need to check if it is Ext2Fs must set `disk_size`.

tools/gen_internal_conf/src/main.rs
lucassong-mh: Done

tools/gen_internal_conf/src/main.rs
lucassong-mh: Done

volcano0dr (approved):
LGTM

PR #1615, opened 2024-09-13
Fix uring ci

ClawSeven:


src/libos/src/net/socket/uring/stream/mod.rs
```diff
@@ -454,8 +454,6 @@ impl<A: Addr, R: Runtime> StreamSocket<A, R> {
     }
 
     fn set_kernel_send_buf_size(&self, buf_size: usize) {
-        // Setting the minimal buf_size to 128 Kbytes
-        let buf_size = (128 * 1024 + 1).max(buf_size);
```
volcano0dr, line 458: Should I consider “+1” ?

volcano0dr (approved):
LGTM

PR #1632, opened 2024-10-29
[libos] Remove uring udp socket redundant msg when returning error

ClawSeven:


src/libos/src/net/socket/uring/datagram/sender.rs
```diff
@@ -169,8 +169,6 @@ impl<A: Addr, R: Runtime> Sender<A, R> {
         let total_copied = msg.copy_buf(bufs)?;
         msg.copy_control(control)?;
 
-        let msghdr_ptr = new_send_req(&mut msg, addr);
```
volcano0dr, line 172: This line should not be deleted.

volcano0dr (approved):
LGTM

PR #1681, opened 2025-04-17
Fix broadcast of Notifier when some observers have been dropped

LeoneChen:


src/libos/src/events/notifier.rs
```diff
@@ -47,21 +47,19 @@ impl<E: Event, F: EventFilter<E>> Notifier<E, F> {
...
     pub fn broadcast(&self, event: &E) {
-        let subscribers = self.subscribers.lock();
-        for subscriber in subscribers.iter() {
-            if let Some(filter) = subscriber.filter.as_ref() {
-                if !filter.filter(event) {
-                    continue;
+        let mut subscribers = self.subscribers.lock();
+        subscribers.retain(|subscriber| {
+            if let Some(observer) = subscriber.observer.upgrade() {
+                if let Some(filter) = subscriber.filter.as_ref() {
```
volcano0dr, line 53: If `filter` is `None`,  also need to call `observer.on_event`.

volcano0dr (approved):
LGTM

PR #1692, opened 2025-09-10
[libos] Implement AIO syscalls

charming-c:


src/libos/src/fs/host_fd.rs
```diff
@@ -41,19 +41,21 @@ impl Drop for HostFd {
...
+pub struct HostFdRegistry {
     set: HashSet<FileDesc>,
+    map: Mutex<HashMap<FileDesc, FileDesc>>,
 }
 
 impl HostFdRegistry {
     pub fn new() -> Self {
         Self {
             set: HashSet::new(),
+            map: Mutex::new(HashMap::new()),
```
elainewu1222: Use rwlock instead. Get fd operations could be much more than update.

src/libos/src/fs/file_table.rs
```diff
@@ -209,6 +223,38 @@ impl FileTable {
         let del_event = FileTableEvent::Del(fd);
         self.notifier.broadcast(&del_event);
     }
+
+    fn try_insert_hostfd(&self, file: &FileRef, fd: FileDesc) {
+        let inode_file = match file.as_inode_file() {
```
elainewu1222:
``` 
let inode_file = if let Ok(f) = file.as_inode_file() {
    f
} else {
    xxxx
}

```

same for other match?

src/libos/src/fs/file_table.rs
```diff
@@ -209,6 +223,38 @@ impl FileTable {
...
+            Some(m) => m,
+            None => {
+                debug!("try insert hostfd: not mnode");
+                return;
+            }
+        };
+
+        let hnode = match mnode.inode.downcast_ref::<HNode>() {
+            Some(h) => h,
+            None => return,
```
elainewu1222: Add log for all error cases

src/libos/src/fs/file_table.rs
```diff
@@ -209,6 +223,38 @@ impl FileTable {
         let del_event = FileTableEvent::Del(fd);
         self.notifier.broadcast(&del_event);
     }
+
+    fn try_insert_hostfd(&self, file: &FileRef, fd: FileDesc) {
```
elainewu1222, line 227:
```
impl HostFd {
    pub fn new(host_fd: FileDesc) -> Self {
        HOST_FD_REGISTRY.lock().unwrap().register(host_fd).unwrap();
        Self(host_fd)
    }
    pub fn to_raw(&self) -> FileDesc {
        self.0
    }
}
```
Why not insert fd->host_fd mapping in HostFd::new?

src/libos/src/fs/host_fd.rs
```diff
@@ -72,4 +74,19 @@ impl HostFdRegistry {
         }
         Ok(())
     }
+
+    pub fn insert_map(&self, key: FileDesc, value: FileDesc) -> Option<FileDesc> {
+        let mut map = self.map.lock();
```
elainewu1222: merge the two line:
```
self.map.lock().insert(key, value)
```

same for other interfaces.

src/libos/src/fs/host_fd.rs
```diff
@@ -41,19 +41,21 @@ impl Drop for HostFd {
...
+    pub static ref HOST_FD_REGISTRY: SgxMutex<HostFdRegistry> =
         { SgxMutex::new(HostFdRegistry::new()) };
 }
 
-/// A registery for host fds to ensure that they are unique.
-struct HostFdRegistry {
+/// A registery for host fds to ensure that they are unique.pub
+pub struct HostFdRegistry {
     set: HashSet<FileDesc>,
+    map: Mutex<HashMap<FileDesc, FileDesc>>,
```
elainewu1222: There is already a lock in HOST_FD_REGISTRY, I guess no lock is required here?

src/libos/src/fs/file_table.rs
charming-c: yes

src/libos/src/fs/file_table.rs
charming-c: OK

src/libos/src/fs/file_table.rs
charming-c: Because Linux AIO only supports file-type file descriptors, its implementation is unified with the file_table in Occlum. If the file descriptor comes from the host OS and is of a file type, it will be inserted into the map.

src/libos/src/fs/host_fd.rs
charming-c: ok

src/libos/src/fs/host_fd.rs
charming-c: You’re right, that was my oversight. No extra lock is needed here. Thanks for catching that!

